PR #7, opened 2020-09-17
refactor: Upgrade indexer framework version

### LINKED ISSUE #1: refactor: Upgrade Indexer for Explorer to use NEAR Indexer Framework of version 0.2.x
There is up to date version on NEAR Indexer Framework that returns `local_receipts`.

We need to use commit hash `8bf9b050e1173aad3d5cdc31de00f3abfb4a0718` in dependency

khorolets:
Also remove the outdated code that handles outcomes, transactions, receipts, access keys, accounts

Closes #1

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,10 +3,7 @@ use num_traits::cast::FromPrimitive;
 use actix;
 use bigdecimal::BigDecimal;
 use clap::derive::Clap;
-use diesel::{
-    prelude::*,
-    dsl,
-};
+use diesel::{dsl, prelude::*};
 #[macro_use]
 extern crate diesel;
 use tokio::sync::mpsc;
@@ -22,7 +19,7 @@ mod configs;
 mod models;
 mod schema;
 
-async fn listen_blocks(mut stream: mpsc::Receiver<near_indexer::BlockResponse>) {
+async fn listen_blocks(mut stream: mpsc::Receiver<near_indexer::StreamerMessage>) {
     let pool = models::establish_connection();
 
     while let Some(block) = stream.recv().await {
@@ -34,7 +31,7 @@ async fn listen_blocks(mut stream: mpsc::Receiver<near_indexer::BlockResponse>) 
             .execute_async(&pool)
             .await
         {
-            Ok(_) => {},
+            Ok(_) => {}
             Err(_) => continue,
         };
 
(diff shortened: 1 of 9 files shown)


src/models/access_keys.rs
```diff
@@ -17,7 +16,9 @@ impl AccessKey {
             account_id,
             public_key: public_key,
             access_key_type: match access_key_type.permission {
-                ref _data @ AccessKeyPermissionView::FunctionCall { .. } => "function_call".to_string(),
+                ref _data @ AccessKeyPermissionView::FunctionCall { .. } => {
```
frol, line 19:
```suggestion
                AccessKeyPermissionView::FunctionCall { .. } => {
```

src/models/accounts.rs
```diff
@@ -12,17 +12,12 @@ pub struct Account {
 }
 
 impl Account {
-    pub fn new(
-        account_id: String,
-        index: i32,
-        receipt_id: String,
-        timestamp: BigDecimal
-    ) -> Self {
+    pub fn new(account_id: String, index: i32, receipt_id: String, timestamp: BigDecimal) -> Self {
```
frol, line 15: This whole `new` implementation does not look useful. You may just as well initialize the `Account` struct in place, and I would prefer to do that over `new` method with positional arguments.

src/models/actions.rs
```diff
@@ -5,7 +5,7 @@ use near_indexer;
 use near_indexer::near_primitives::views::ActionView;
 
 use crate::schema;
-use schema::{actions};
+use schema::actions;
```
frol, line 8: They are not in use. Please, address all the warnings and clippy messages.

src/models/receipts.rs
```diff
@@ -36,39 +36,45 @@ pub struct ReceiptAction {
 }
 
 #[derive(Insertable)]
-#[table_name="actions_input_data"]
+#[table_name = "actions_input_data"]
```
frol, line 39: Should these be `receipt_actions_input_data`? Do you even need to name them explicitly? (I assumed that those are autogenerated)

src/models/receipts.rs
```diff
@@ -36,39 +36,45 @@ pub struct ReceiptAction {
...
             receipt_id: receipt_view.receipt_id.to_string(),
             predecessor_id: Some(receipt_view.predecessor_id.to_string()),
             receiver_id: Some(receipt_view.receiver_id.to_string()),
             status: "empty".to_string(),
             type_: match &receipt_view.receipt {
-                ref _entity @ near_indexer::near_primitives::views::ReceiptEnumView::Action { .. } => Some("action".to_string()),
-                ref _entity @ near_indexer::near_primitives::views::ReceiptEnumView::Data { .. } => Some("data".to_string()),
+                ref
+                _entity
+                @
```
frol, line 63: why do you need these _entity mapping?

src/main.rs
```diff
@@ -22,277 +16,42 @@ mod configs;
...
-                                    },
-                                    _ => {},
-                                };
-                            }
-                    }
-                }
-
+            Ok(_) => {}
+            Err(_) => {
+                eprintln!("Unable to save chunk, skipping");
```
frol, line 49: loop and retry forever with error messages

frol (approved):
If it is essentially just an upgrade, LGTM. Let's merge it.

PR #12, opened 2020-10-06
feat: index accounts from receipt execution outcomes

### LINKED ISSUE #6: refactor: Extract Accounts from receipts
Similar to #5 we want to extract Accounts

khorolets:
*Resolves #6*

### DIFF
--- a/src/db_adapters/accounts.rs
+++ b/src/db_adapters/accounts.rs
@@ -92,7 +92,6 @@ async fn remove_accounts(
     pool: &Pool<ConnectionManager<PgConnection>>,
     outcomes: &[near_indexer::IndexerExecutionOutcomeWithReceipt],
 ) {
-    eprintln!("{:#?}", &outcomes);
     let accounts_to_delete: Vec<(String, String)> = outcomes
         .iter()
         .filter_map(|outcome_with_receipt| {
@@ -106,12 +105,10 @@ async fn remove_accounts(
                                     ..
                                 } = action
                                 {
-                                    Some(
-                                        (
-                                            (&receipt.receiver_id).to_string(),
-                                            receipt.receipt_id.to_string(),
-                                        )
-                                    )
+                                    Some((
+                                        (&receipt.receiver_id).to_string(),
+                                        receipt.receipt_id.to_string(),
+                                    ))
                                 } else {
                                     None
                                 }
@@ -128,15 +125,11 @@ async fn remove_accounts(
         .flatten()
         .collect();
 
-    eprintln!("ACCOUNT_TO_DELETE:\n {:#?}", &accounts_to_delete);
-
     for (account_id, deleted_by_receipt_id) in accounts_to_delete {
         loop {
             match diesel::update(schema::accounts::table)
                 .filter(schema::accounts::dsl::account_id.eq(account_id.clone()))
-                .set(
-                    schema::accounts::dsl::deleted_by_receipt_id.eq(deleted_by_receipt_id.clone())
-                )
+                .set(schema::accounts::dsl::deleted_by_receipt_id.eq(deleted_by_receipt_id.clone()))
                 .execute_async(&pool)
                 .await
             {
--- a/src/db_adapters/accounts.rs
+++ b/src/db_adapters/accounts.rs
@@ -8,13 +8,17 @@ use near_indexer::near_primitives;
 
 use crate::models;
 use crate::schema;
+use diesel::pg::upsert::excluded;
 
 /// Saves new Accounts to database or deletes the ones should be deleted
 pub(crate) async fn handle_accounts(
     pool: &Pool<ConnectionManager<PgConnection>>,
     outcomes: &near_indexer::ExecutionOutcomesWithReceipts,
 ) {
-    let successful_receipts_with_actions: Vec<(&near_primitives::views::ReceiptView, &Vec<near_primitives::views::ActionView>)> = outcomes
+    let successful_receipts_with_actions: Vec<(
+        &near_primitives::views::ReceiptView,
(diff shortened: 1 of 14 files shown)


src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,157 @@
...
+                            .collect();
+                        Some(accounts)
+                    }
+                    _ => None,
+                }
+            } else {
+                None
+            }
+        })
+        .flatten()
```
frol, line 66: A single receipt can only create a single account (having multiple `CreateAccount` actions in a single receipt is pointless):

```suggestion
        .filter_map(|outcome_with_receipt| outcome_with_receipt.receipt)
        .filter_map(|receipt| if let near_primitives::views::ReceiptEnumView::Action { actions, .. } = receipt.receipt {
            Some((receipt, actions))
        } else {
            None
        })
        .filter_map(|(receipt, actions)| actions
            .iter()
            .filter(|action| matches!(action, near_primitives::views::ActionView::CreateAccount))
            .map(|_create_account_action| models::accounts::Account::new(
                receipt.receiver_id.to_string(),
                &receipt.receipt_id,
            )
            .next()
        )
```

src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,157 @@
...
+        .collect();
+
+    eprintln!("ACCOUNT_TO_DELETE:\n {:#?}", &accounts_to_delete);
+
+    for (account_id, deleted_by_receipt_id) in accounts_to_delete {
```
frol, line 128: This is quite inefficient, but DeleteAccounts are rare, so we can live with it as is. Just FYI, you can update multiple records to distict values in a single query: https://stackoverflow.com/a/18799497/1178806

src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+async fn remove_accounts(
+    pool: &Pool<ConnectionManager<PgConnection>>,
+    outcomes: &[near_indexer::IndexerExecutionOutcomeWithReceipt],
+) {
+    eprintln!("{:#?}", &outcomes);
+    let accounts_to_delete: Vec<(String, String)> = outcomes
+        .iter()
+        .filter_map(|outcome_with_receipt| {
```
frol, line 97: Apply the approach I posted above to this chunk as well

src/db_adapters/execution_outcomes.rs
```diff
@@ -44,18 +44,26 @@ pub(crate) async fn store_execution_outcomes(
...
-            |(index, receipt_id)| models::execution_outcomes::ExecutionOutcomeReceipt {
-                execution_outcome_receipt_id: outcome.id.to_string(),
-                index: index as i32,
-                receipt_id: receipt_id.to_string(),
-            },
-        ));
+        outcome_receipt_models.extend(
+            outcome
+                .execution_outcome
+                .outcome
```
frol, line 55: How deep is the rabbit hole of outcome.outcome.outcome? :smile:

src/models/accounts.rs
```diff
@@ -1,23 +1,22 @@
-use bigdecimal::BigDecimal;
-
 use crate::schema;
 use schema::accounts;
 
-#[derive(Insertable)]
+#[derive(Insertable, Queryable, Debug, Clone)]
```
frol, line 4: Why do we need `Queryable` here?

src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,150 @@
...
+                    near_primitives::views::ReceiptEnumView::Action { actions, .. } => {
+                        let accounts: Vec<(String, String)> = actions
+                            .iter()
+                            .filter_map(|action| {
+                                if let near_primitives::views::ActionView::DeleteAccount {
+                                    ..
+                                } = action
+                                {
+                                    Some((
+                                        (&receipt.receiver_id).to_string(),
```
frol, line 109:
```suggestion
                                        receipt.receiver_id.to_string(),
```

src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,150 @@
...
+                None
+            }
+        })
+        .flatten()
+        .collect();
+
+    loop {
+        match diesel::insert_into(schema::accounts::table)
+            .values(accounts_to_create.clone())
+            .on_conflict_do_nothing()
```
frol, line 72: If there is an existing account, we should update its `created_by_receipt_id`. (e.g. in case CreateAccount -> DeleteAccount -> CreateAccount we should arrive with Account holding the last receipt_id in `created_by_receipt_id` and `deleted_by_receipt_id` should be NULL, I suppose)

frol (changes requested):

src/db_adapters/accounts.rs
```diff
@@ -0,0 +1,157 @@
...
+                _ => false,
+            }
+        })
+        .cloned()
+        .collect();
+
+    let store_accounts_future = store_accounts(&pool, &successful_outcomes);
+    let remove_accounts_future = remove_accounts(&pool, &successful_outcomes);
+
+    join!(store_accounts_future, remove_accounts_future);
```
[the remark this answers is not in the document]
khorolets: Outcomes we have from streamer don't persist the correct order anyway. What do you suggest here? Await them one by one? We don't have correct order in the first place and we totally don't have it after filtering accounts.

src/db_adapters/accounts.rs
khorolets: Just to clarify. Pointless or impossible?

frol: I don't know about impossibility, but it is clearly pointless.

frol (approved):

PR #16, opened 2020-10-09
feat: Add command arg to choose SyncMode for Streamer

### LINKED ISSUE #15: feat: Add command arg to choose SyncMode for Streamer
Currently streamers SyncMode is hardcoded and can be changed before compilation, but we want to be able to pass it like an args. This will allow us to run the same binary on different machines with different settings (like one for real-time indexing and the other one for historical data)

khorolets:
closes #15 

Command to run Indexer for Explorer updated

Before
```
indexer-binary --home-dir ~/.near/testnet run
```

After

```
indexer-binary --home-dir ~/.near/testnet run --sync-mode from-interruption
indexer-binary --home-dir ~/.near/testnet run --sync-mode last-synced
indexer-binary --home-dir ~/.near/testnet run --sync-mode block --height <block_height>
```

However the "before" command still works and `--sync-mode` default will be `from-interruption`

### DIFF
--- a/src/configs.rs
+++ b/src/configs.rs
@@ -1,3 +1,6 @@
+use std::convert::TryFrom;
+use std::str::FromStr;
+
 use clap::Clap;
 
 /// NEAR Indexer for Explorer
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -8,6 +8,7 @@ use tracing::info;
 use tracing_subscriber::EnvFilter;
 
 use crate::configs::{Opts, SubCommand};
+use std::convert::TryInto;
```
frol, line 11: Nit: order the imports

src/configs.rs
```diff
@@ -15,11 +18,60 @@ pub(crate) struct Opts {
...
+    Run(RunArgs),
     /// Initialize necessary configs
     Init(InitConfigArgs),
 }
 
+#[derive(Clap, Debug)]
+pub(crate) struct RunArgs {
+    /// streamer SyncMode. Possible values: from-interruption, last-synced and block. Default: from-interruption
+    #[clap(long, default_value = "from-interruption")]
+    pub sync_mode: SyncMode,
```
frol, line 30: It is fine to have it this way, but if you want to learn how we can improve it, consider using sync modes as separate subcommands where only SyncFromBlock subcommand will have the block height (and it will be a required/non-Option field)

frol (approved):

frol (approved):

PR #17, opened 2020-10-14
feat: Add accounts from genesis on start

### LINKED ISSUE #13: feat: Add accounts from genesis on Indexer start
In order to add all the accounts from genesis into database in indexer we need to introduce genesis handler, which will add all the accounts from genesis file to database on Indexer start

khorolets:
Closes #13 

Introduce additional parameter to `run` command `--store-genesis` (couldn't find better name) to include handling of genesis file on start of indexer.

Currently if activated will run the function that will walk over `genesis.records` to add records in db with accounts.

~**Note** If account is already present in db this handler will update `created_by_receipt_id` and `deleted_by_receipt_id` by setting them to `NULL` (assuming account is created in genesis and haven't been deleted), one should carefully use it on Indexer restart.~

@frol let me know what do you think about the behaviour described above.

### DIFF
--- a/src/db_adapters/accounts.rs
+++ b/src/db_adapters/accounts.rs
@@ -2,7 +2,7 @@ use diesel::r2d2::{ConnectionManager, Pool};
 use diesel::{ExpressionMethods, PgConnection};
 use futures::join;
 use tokio_diesel::AsyncRunQueryDsl;
-use tracing::error;
+use tracing::{error, info};
 
 use near_indexer::near_primitives;
 
@@ -65,7 +65,7 @@ async fn store_accounts(
                 .map(|_create_account_action| {
                     models::accounts::Account::new(
                         receipt.receiver_id.to_string(),
-                        &receipt.receipt_id,
+                        Some(&receipt.receipt_id),
                     )
                 })
                 .next()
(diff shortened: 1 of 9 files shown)


src/db_adapters/accounts.rs
```diff
@@ -141,3 +141,62 @@ async fn remove_accounts(
...
+    info!(
+        target: crate::INDEXER_FOR_EXPLORER,
+        "Adding/updating accounts from genesis..."
+    );
+    let pool = crate::models::establish_connection();
+
+    let accounts_models: Vec<models::accounts::Account> = near_config
+        .genesis
+        .records
+        .0
```
frol, line 155:
```suggestion
        .as_ref()
```

src/db_adapters/accounts.rs
```diff
@@ -141,3 +141,62 @@ async fn remove_accounts(
...
+        match diesel::insert_into(schema::accounts::table)
+            .values(accounts_models.clone())
+            .on_conflict(schema::accounts::dsl::account_id)
+            .do_update()
+            .set((
+                schema::accounts::dsl::created_by_receipt_id
+                    .eq(excluded(schema::accounts::dsl::created_by_receipt_id)),
+                schema::accounts::dsl::deleted_by_receipt_id
+                    .eq(excluded(schema::accounts::dsl::deleted_by_receipt_id)),
+            ))
```
frol, line 180: I think we should do nothing with the account that was created in response to some execution outcome (which happens after genesis). We may even want to only dump genesis if the database is completely blank since it does not make a lot of sense to mix some stale state with the fresh genesis dump :thinking:

src/models/accounts.rs
```diff
@@ -4,18 +4,21 @@ use schema::accounts;
...
-        created_by_receipt_id: &near_indexer::near_primitives::hash::CryptoHash,
+        created_by_receipt_id: Option<&near_indexer::near_primitives::hash::CryptoHash>,
     ) -> Self {
         Self {
             account_id,
-            created_by_receipt_id: created_by_receipt_id.to_string(),
+            created_by_receipt_id: match created_by_receipt_id {
+                Some(receipt_id) => Some(receipt_id.to_string()),
+                None => None,
+            },
```
frol, line 21: FYI:

```rust
            created_by_receipt_id: created_by_receipt_id.cloned(),
```

src/models/accounts.rs
```diff
@@ -4,18 +4,21 @@ use schema::accounts;
...
-    pub created_by_receipt_id: String,
+    pub created_by_receipt_id: Option<String>,
     pub deleted_by_receipt_id: Option<String>,
 }
 
 impl Account {
     pub fn new(
         account_id: String,
-        created_by_receipt_id: &near_indexer::near_primitives::hash::CryptoHash,
+        created_by_receipt_id: Option<&near_indexer::near_primitives::hash::CryptoHash>,
```
frol, line 14: I suggest we have two methods: `new_from_genesis` (only `account_id` is required) and `new_from_receipt` (`account_id` and `created_by_receipt_id` is required).

src/models/accounts.rs
```diff
@@ -4,18 +4,21 @@ use schema::accounts;
 #[derive(Insertable, Debug, Clone)]
 pub struct Account {
     pub account_id: String,
-    pub created_by_receipt_id: String,
+    pub created_by_receipt_id: Option<String>,
```
frol, line 7: Do we really need the fields to be `pub`?

src/db_adapters/accounts.rs
```diff
@@ -141,3 +141,62 @@ async fn remove_accounts(
...
+                Some(models::accounts::Account::new(account_id.to_string(), None))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    loop {
+        match diesel::insert_into(schema::accounts::table)
+            .values(accounts_models.clone())
```
frol, line 172: There are 213596 accounts in testnet genesis. Let's split this into batches as we do in Indexer for Wallet, I guess.

frol (changes requested):

frol (approved):

PR #20, opened 2020-10-16
feat: Extract AccessKeys from receipts (like we do in Indexer forExtract AccessKeys from receipts (like we do in Indexer for Wallet)

### LINKED ISSUE #5: refactor: Extract AccessKeys from receipts (like we do in Indexer for Wallet)
We want to store `AccessKey` actions separately to query them in future like we do in Indexer for Wallet. In addition to Indexer for Wallet logic we also store receipts as is (see #2). That's why we want to introduce some sort of handler we can attach to receipts processing providing additional logic.

Need to discuss with @frol before starting

khorolets:
Closes #5

### DIFF
--- a/src/db_adapters/access_keys.rs
+++ b/src/db_adapters/access_keys.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 
+use diesel::pg::upsert::excluded;
 use diesel::r2d2::{ConnectionManager, Pool};
 use diesel::{ExpressionMethods, PgConnection, QueryDsl};
+use futures::join;
 use tokio_diesel::AsyncRunQueryDsl;
 use tracing::error;
 
@@ -9,7 +11,6 @@ use near_indexer::near_primitives;
 
 use crate::models;
 use crate::schema;
-use diesel::pg::upsert::excluded;
 
 pub(crate) async fn handle_access_keys(
     pool: &Pool<ConnectionManager<PgConnection>>,
@@ -36,42 +37,29 @@ pub(crate) async fn handle_access_keys(
                         public_key,
                         access_key,
                     } => {
-                        if let Some(existing_record) = access_keys
-                            .get_mut(&(public_key.to_string(), receipt.receiver_id.to_string()))
-                        {
-                            existing_record.created_by_receipt_id =
-                                Some(receipt.receipt_id.to_string());
-                            existing_record.deleted_by_receipt_id = None;
-                        } else {
-                            access_keys.insert(
-                                (public_key.to_string(), receipt.receiver_id.to_string()),
-                                models::access_keys::AccessKey::from_action_view(
-                                    public_key,
-                                    &receipt.receiver_id,
-                                    access_key,
-                                    &receipt.receipt_id,
-                                ),
-                            );
-                        }
+                        access_keys.insert(
+                            (public_key.to_string(), receipt.receiver_id.to_string()),
+                            models::access_keys::AccessKey::from_action_view(
+                                public_key,
+                                &receipt.receiver_id,
+                                access_key,
+                                &receipt.receipt_id,
+                            ),
+                        );
                     }
                     near_primitives::views::ActionView::DeleteKey { public_key } => {
-                        if let Some(existing_record) = access_keys
-                            .get_mut(&(public_key.to_string(), receipt.receiver_id.to_string()))
-                        {
-                            existing_record.deleted_by_receipt_id =
-                                Some(receipt.receipt_id.to_string());
-                        } else {
-                            access_keys.insert(
-                                (public_key.to_string(), receipt.receiver_id.to_string()),
-                                models::access_keys::AccessKey {
-                                    public_key: public_key.to_string(),
-                                    account_id: receipt.receiver_id.to_string(),
-                                    created_by_receipt_id: None,
-                                    deleted_by_receipt_id: Some(receipt.receipt_id.to_string()),
-                                    permission: models::enums::AccessKeyPermission::NotApplicable,
-                                },
-                            );
-                        }
+                        access_keys.insert(
+                            (public_key.to_string(), receipt.receiver_id.to_string()),
+                            models::access_keys::AccessKey {
+                                public_key: public_key.to_string(),
+                                account_id: receipt.receiver_id.to_string(),
+                                created_by_receipt_id: None,
+                                deleted_by_receipt_id: Some(receipt.receipt_id.to_string()),
+                                // this is a workaround to avoid additional struct with optional field
+                                // permission_kind is not supposed to change on delete action
+                                permission_kind: models::enums::AccessKeyPermission::FullAccess,
+                            },
+                        );
                     }
                     _ => continue,
                 }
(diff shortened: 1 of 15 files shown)


src/db_adapters/access_keys.rs
```diff
@@ -0,0 +1,141 @@
...
+                            access_keys.insert(
+                                (public_key.to_string(), receipt.receiver_id.to_string()),
+                                models::access_keys::AccessKey::from_action_view(
+                                    public_key,
+                                    &receipt.receiver_id,
+                                    access_key,
+                                    &receipt.receipt_id,
+                                ),
+                            );
+                        }
```
frol, line 55: We don't want to reuse any of the previous fields if we see AddKey (imagine a case of AddKey FullAccess, DeleteKey, AddKey FunctionCall):

```suggestion
                        access_keys
                            .insert(
                                (public_key.to_string(), receipt.receiver_id.to_string()),
                                models::access_keys::AccessKey::from_action_view(
                                    public_key,
                                    &receipt.receiver_id,
                                    access_key,
                                    &receipt.receipt_id,
                                )
                            )
```

src/db_adapters/access_keys.rs
```diff
@@ -0,0 +1,141 @@
...
+                                Some(receipt.receipt_id.to_string());
+                        } else {
+                            access_keys.insert(
+                                (public_key.to_string(), receipt.receiver_id.to_string()),
+                                models::access_keys::AccessKey {
+                                    public_key: public_key.to_string(),
+                                    account_id: receipt.receiver_id.to_string(),
+                                    created_by_receipt_id: None,
+                                    deleted_by_receipt_id: Some(receipt.receipt_id.to_string()),
+                                    permission: models::enums::AccessKeyPermission::NotApplicable,
```
frol, line 71: Well, I was wrong with my suggestion for using the model itself. It will be cleaner if we follow your approach of having a custom structure (we at least want to have `permission_kind` optional since it does not make sense to have NOT_APPLICABLE in the database, alternatively, as a cruel workaround we can set it to FULL_ACCESS here)

src/db_adapters/access_keys.rs
```diff
@@ -0,0 +1,141 @@
...
+                    near_primitives::views::ActionView::DeleteKey { public_key } => {
+                        if let Some(existing_record) = access_keys
+                            .get_mut(&(public_key.to_string(), receipt.receiver_id.to_string()))
+                        {
+                            existing_record.deleted_by_receipt_id =
+                                Some(receipt.receipt_id.to_string());
+                        } else {
+                            access_keys.insert(
+                                (public_key.to_string(), receipt.receiver_id.to_string()),
+                                models::access_keys::AccessKey {
```
frol, line 66: Here is how you can express it:

```rust
access_keys
    .entry((public_key.to_string(), receipt.receiver_id.to_string()))
    .and_modify(|existing_access_key| {
        existing_access_key.deleted_by_receipt_id = Some(receipt.receipt_id.to_string());
    })
    .or_insert_with(||
        models::access_key::AccessKey { ... }
    )
```

src/db_adapters/access_keys.rs
```diff
@@ -0,0 +1,141 @@
...
+                            );
+                        }
+                    }
+                    _ => continue,
+                }
+            }
+        }
+    }
+
+    let mut access_keys_to_insert: Vec<models::access_keys::AccessKey> = vec![];
```
frol, line 82: You can [`partition`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.partition) the `access_keys`, and run the queries in parallel (using FuturesUnordered for the updates + async block + `join!(updates_block, inserts_block)`)

src/db_adapters/access_keys.rs
```diff
@@ -0,0 +1,141 @@
...
+    loop {
+        match diesel::insert_into(schema::access_keys::table)
+            .values(access_keys_to_insert.clone())
+            .on_conflict((schema::access_keys::dsl::public_key, schema::access_keys::dsl::account_id))
+            .do_update()
+            .set((
+                schema::access_keys::dsl::created_by_receipt_id
+                    .eq(excluded(schema::access_keys::dsl::created_by_receipt_id)),
+                schema::access_keys::dsl::deleted_by_receipt_id
+                    .eq(excluded(schema::access_keys::dsl::deleted_by_receipt_id)),
```
frol, line 124: We should also override `permission_kind`. Is there a way to just say that we want to override all the fields?

src/main.rs
```diff
@@ -66,6 +66,12 @@ async fn handle_message(
     // Accounts
     db_adapters::accounts::handle_accounts(&pool, &streamer_message.receipt_execution_outcomes)
         .await;
+    // AccessKeys
+    db_adapters::access_keys::handle_access_keys(
+        &pool,
+        &streamer_message.receipt_execution_outcomes,
+    )
+    .await;
```
frol, line 74: I guess we can already parallelize some of the handlers (execution outcomes, accounts, access keys are independent)

src/models/enums.rs
```diff
@@ -58,3 +58,42 @@ impl From<near_indexer::near_primitives::views::ExecutionStatusView> for Executi
         }
     }
 }
+
+#[derive(Debug, DbEnum, Clone)]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+#[DieselType = "Access_key_permission_type"]
+#[PgType = "access_key_permission_type"]
+pub enum AccessKeyPermission {
```
frol, line 66:
```suggestion
#[PgType = "access_key_permission_kind"]
pub enum AccessKeyPermissionKind {
```

src/models/enums.rs
```diff
@@ -58,3 +58,42 @@ impl From<near_indexer::near_primitives::views::ExecutionStatusView> for Executi
...
+    /// Used only with AccessKeyAction::Delete
+    NotApplicable,
```
frol, line 68:
```suggestion
```

src/models/access_keys.rs
```diff
@@ -1,26 +1,81 @@
...
-    pub access_key_type: String,
+    pub account_id: String,
+    pub created_by_receipt_id: Option<String>,
+    pub deleted_by_receipt_id: Option<String>,
+    pub permission: AccessKeyPermission,
 }
 
 impl AccessKey {
-    pub fn new(account_id: String, public_key: String, access_key_type: &AccessKeyView) -> Self {
+    pub fn from_receipt_view(
```
frol, line 17: Where do we use it? Seems to be a dead code for me

src/models/enums.rs
```diff
@@ -58,3 +58,42 @@ impl From<near_indexer::near_primitives::views::ExecutionStatusView> for Executi
...
+                ..
+            } => Self::FunctionCall,
+            near_indexer::near_primitives::views::AccessKeyPermissionView::FullAccess => {
+                Self::FullAccess
+            }
+        }
+    }
+}
+
+impl From<&near_indexer::near_primitives::account::AccessKeyPermission> for AccessKeyPermission {
```
frol, line 88: Where do we get it from (I mean non-View struct)? :thinking:

frol (changes requested):

src/db_adapters/access_keys.rs
```diff
@@ -36,53 +37,46 @@ pub(crate) async fn handle_access_keys(
...
-                                    permission: models::enums::AccessKeyPermission::NotApplicable,
-                                },
-                            );
-                        }
+                        access_keys.insert(
+                            (public_key.to_string(), receipt.receiver_id.to_string()),
+                            models::access_keys::AccessKey {
+                                public_key: public_key.to_string(),
+                                account_id: receipt.receiver_id.to_string(),
+                                created_by_receipt_id: None,
```
frol, line 56: We want to preserve the `created_by_receipt_id` if we had this access_key just added (thus I used `and_modify` in my comment)

src/db_adapters/access_keys.rs
```diff
@@ -36,53 +37,46 @@ pub(crate) async fn handle_access_keys(
...
+                        );
                     }
                     _ => continue,
                 }
             }
         }
     }
 
-    let mut access_keys_to_insert: Vec<models::access_keys::AccessKey> = vec![];
+    let (access_keys_to_insert, access_keys_to_delete): (
```
frol, line 70:
```suggestion
    let (access_keys_to_insert, access_keys_to_update): (
```

frol (changes requested):

frol (approved):

PR #23, opened 2020-10-20
feat: Improve Access Key serialization in AddKey actions

### LINKED ISSUE #19: Improve Access Key serialization in AddKey actions
Currently, we store the serialized action argument with AccessKeyView as

```json
{
  "access_key": {
    "nonce": 0,
    "permission": "FullAccess"
  },
  "public_key": "ed25519:4Rx5XC1Nu1q4TvNatTpokM8mrWGLiCYtHccWpu3VaDJd"
}
```

and

```json
{
  "access_key": {
    "nonce": 0,
    "permission": {
      "FunctionCall": {
        "allowance": "100500",
        "receiver_id": "contract-name",
        "method_names": ["..."]
      }
    }
  },
  "public_key": "ed25519:4Rx5XC1Nu1q4TvNatTpokM8mrWGLiCYtHccWpu3VaDJd"
}
```

Notice that the `permissions` value can be either a string or an object. We should have a unified structure using [serde tagger enum representation](https://serde.rs/enum-representations.html) (and I also would prefer to have the const strings to be in SCREAMING_SNAKE_CASE):

```json
{
  "access_key": {
    "nonce": 0,
    "permission_kind": "FULL_ACCESS"
  },
  "public_key": "ed25519:4Rx5XC1Nu1q4TvNatTpokM8mrWGLiCYtHccWpu3VaDJd"
}
```

and

```json
{
  "access_key": {
    "nonce": 0,
    "permission_kind": "FUNCTION_CALL",
    "permission_details": {
      "allowance": "100500",
      "receiver_id": "contract-name",
      "method_names": ["..."]
    }
  },
  "public_key": "ed25519:4Rx5XC1Nu1q4TvNatTpokM8mrWGLiCYtHccWpu3VaDJd"
}
```

The following config should do the trick:

```
#[serde(tag = "permission_kind", content = "permission_details", rename_all = "SCREAMIING_SNAKE_CASE")]
```

, but we cannot apply it directly to the `PermissionView` enum, so we will need to do some code duplication and implement `From` trait to convert the nearcore primitive structure into our Indexer custom implementation.

khorolets:
Closes #19


src/models/serializer.rs
```diff
@@ -0,0 +1,46 @@
+use serde::{Deserialize, Serialize};
+use near_indexer::near_primitives::serialize::option_u128_dec_format;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub(crate) struct AccessKeyView {
```
frol, line 5: Let's add some comments describing the goal we try to achive.

src/models/serializer.rs
```diff
@@ -0,0 +1,46 @@
...
+        Self {
+            nonce: access_key_view.nonce,
+            permission: access_key_view.permission.clone().into(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(tag = "permission_kind", content = "permission_details", rename_all = "SCREAMING_SNAKE_CASE")]
+pub(crate) enum AccessKeyPermissionView {
```
frol, line 21: same

src/models/mod.rs
```diff
@@ -84,7 +87,7 @@ pub(crate) fn extract_action_type_and_value_from_action_view(
             ActionKind::AddKey,
             json!({
                 "public_key": public_key,
-                "access_key": access_key,
+                "access_key": AccessKeyView::from(access_key),
```
frol, line 90: This whole `match` in `extract_action_type_and_value_from_action_view` seems to belong to `serializers`. Yet, I think we can get away with it as is for now. I only suggest you use `AccessKeyView` by its path as the name is colliding with the nearcore struct name:

```suggestion
                "access_key": self::serializers::AccessKeyView::from(access_key),
```

P.S. I suggest we call the module with `s` at the end (`serializers`).

frol (approved):

PR #25, opened 2020-10-26
refacor: Strict mode

### LINKED ISSUE #24: Strict mode
We are running Indexer for Explorer for testnet, started from it's genesis. We have a logic to skip receipts we don't have parent transaction for. While it is possible for running from some point after genesis it shouldn't be possible while running from genesis, there shouldn't be any missing transactions.

We suppose this is happening because we are spawning the handler and occurring in situation where handler which should add transaction hasn't completed yet (hasn't stored transaction to database) but the newer handler is already trying to add receipt that is child for that transaction. And we are skipping the receipt, which is wrong behaviour in this case.

Also we're planning to run a couple of indexer nodes, at least one to index from genesis and another one to index in real-time. So adding retry logic to those receipts is not the solution we're looking for. 

We want to add some sort of strict mode (enabled by default) in which retry logic for receipts will be enabled and could be disabled if strict mode is off.

We came up with the idea to introduce additional command to `run` that will disable strict mode for specific amount of blocks (for the case when we run one node from genesis and another one like real-time to be able to continue but not stuck on such receipts)

So it can look like

```
--allow-missing-relations-in-first-blocks 1000
```

khorolets:
Closes #24

### DIFF
--- a/src/db_adapters/receipts.rs
+++ b/src/db_adapters/receipts.rs
@@ -69,10 +69,9 @@ pub(crate) async fn store_receipts(
 /// Looks for already created parent transaction hash for given receipts
 async fn find_tx_hashes_for_receipts(
     pool: &Pool<ConnectionManager<PgConnection>>,
-    receipt_ids: Vec<String>,
+    mut receipt_ids: Vec<String>,
     strict_mode: bool,
 ) -> HashMap<String, String> {
-    let mut receipts = receipt_ids.clone();
     let mut tx_hashes_for_receipts: HashMap<String, String> = HashMap::new();
 
     let mut retries_left: u8 = 10; // retry at least times even in no-strict mode to avoid data loss
@@ -86,7 +85,8 @@ async fn find_tx_hashes_for_receipts(
                     ),
                 )
                 .filter(
-                    schema::execution_outcome_receipts::dsl::receipt_id.eq(any(receipts.clone())),
+                    schema::execution_outcome_receipts::dsl::receipt_id
+                        .eq(any(receipt_ids.clone())),
                 )
                 .select((
                     schema::execution_outcome_receipts::dsl::receipt_id,
@@ -110,17 +110,18 @@ async fn find_tx_hashes_for_receipts(
             }
         };
 
+        let found_hashes_len = tx_hashes_for_receipts_via_outcomes.len();
         tx_hashes_for_receipts.extend(tx_hashes_for_receipts_via_outcomes);
 
-        if tx_hashes_for_receipts.len() == receipt_ids.len() {
+        if found_hashes_len == receipt_ids.len() {
             break;
         }
 
-        receipts.retain(|r| tx_hashes_for_receipts.contains_key(r.as_str()));
+        receipt_ids.retain(|r| tx_hashes_for_receipts.contains_key(r.as_str()));
 
         let tx_hashes_for_receipt_via_transactions: Vec<(String, String)> = loop {
             match schema::transactions::table
-                .filter(schema::transactions::dsl::receipt_id.eq(any(receipts.clone())))
+                .filter(schema::transactions::dsl::receipt_id.eq(any(receipt_ids.clone())))
                 .select((
                     schema::transactions::dsl::receipt_id,
                     schema::transactions::dsl::transaction_hash,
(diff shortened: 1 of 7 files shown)


src/db_adapters/receipts.rs
```diff
@@ -17,10 +17,98 @@ pub(crate) async fn store_receipts(
     pool: &Pool<ConnectionManager<PgConnection>>,
     receipts: Vec<&near_indexer::near_primitives::views::ReceiptView>,
     block_hash: &str,
+    no_strict_mode: bool,
```
frol, line 20: Let's not use `no_` for `bool` as it creates double-negative in case of `no_strict_mode = false`

src/db_adapters/receipts.rs
```diff
@@ -17,10 +17,98 @@ pub(crate) async fn store_receipts(
...
+                    r,
+                    block_hash,
+                    transaction_hash,
+                ))
+            } else {
+                warn!(
+                    target: crate::INDEXER_FOR_EXPLORER,
+                    "Skipping Receipt {} as we can't find parent Transaction for it.",
+                    r.receipt_id.to_string()
+                );
```
frol, line 47: Let's report these as a single warning message in the `else` branch of the `strict_mode` check.

src/db_adapters/receipts.rs
```diff
@@ -17,10 +17,98 @@ pub(crate) async fn store_receipts(
...
+
+    if !no_strict_mode {
+        let receipt_models_ = loop {
+            let mut models: Vec<models::receipts::Receipt> = vec![];
+            let mut new_skipping_receipt_ids =
+                std::collections::HashSet::<near_indexer::near_primitives::hash::CryptoHash>::new();
+            if skipping_receipt_ids.is_empty() {
+                break models;
+            } else {
+                let tx_hashes_for_receipts = find_tx_hashes_for_receipts(
```
frol, line 62: `warn` here that we are going to fetch the transaction hashes again.

src/db_adapters/receipts.rs
```diff
@@ -17,10 +17,98 @@ pub(crate) async fn store_receipts(
...
+        })
+        .collect();
+
+    if !no_strict_mode {
+        let receipt_models_ = loop {
+            let mut models: Vec<models::receipts::Receipt> = vec![];
+            let mut new_skipping_receipt_ids =
+                std::collections::HashSet::<near_indexer::near_primitives::hash::CryptoHash>::new();
+            if skipping_receipt_ids.is_empty() {
+                break models;
```
frol, line 60: Given we initiate the `models` vector on every iteration, it seems that it will always be empty here. Also, we need to make sure we don't mess up with the `index_in_chunk` here.

frol (commented):
I have not finished the review, but there are some preliminary comments.

src/db_adapters/receipts.rs
```diff
@@ -17,10 +17,105 @@ pub(crate) async fn store_receipts(
...
+
+    let process_receipt_actions_future = store_receipt_actions(&pool, action_receipts);
+
+    let process_receipt_data_future = store_receipt_data(&pool, data_receipts);
+
+    join!(process_receipt_actions_future, process_receipt_data_future);
+}
+
+/// Looks for already created parent transaction hash for given receipts
+async fn find_tx_hashes_for_receipts(
```
frol, line 115: I suggest this function takes all the hassle about the strict mode, so you collect all the transaction hashes first and then you create all the receipt models in a single go.

frol (changes requested):

src/db_adapters/receipts.rs
```diff
@@ -129,6 +66,105 @@ pub(crate) async fn store_receipts(
     join!(process_receipt_actions_future, process_receipt_data_future);
 }
 
+/// Looks for already created parent transaction hash for given receipts
+async fn find_tx_hashes_for_receipts(
+    pool: &Pool<ConnectionManager<PgConnection>>,
+    receipt_ids: Vec<String>,
+    strict_mode: bool,
+) -> HashMap<String, String> {
+    let mut receipts = receipt_ids.clone();
```
frol, line 75:
```suggestion
    mut receipt_ids: Vec<String>,
    strict_mode: bool,
) -> HashMap<String, String> {
```

Fix `receipts` -> `receipt_ids` below.

src/db_adapters/receipts.rs
```diff
@@ -129,6 +66,105 @@ pub(crate) async fn store_receipts(
...
+
+        if tx_hashes_for_receipts.len() == receipt_ids.len() {
+            break;
+        }
+
+        receipts.retain(|r| tx_hashes_for_receipts.contains_key(r.as_str()));
+
+        if !strict_mode {
+            if retries_left > 0 {
+                retries_left = retries_left.saturating_sub(1);
```
frol, line 156:
```suggestion
                retries_left -= 1;
```

There is no need for `saturating_sub` since you explicitly check for `> 0`

frol (approved):

frol (approved):

PR #39, opened 2020-11-05
fix: Fix/missing parent for receipt

### LINKED ISSUE #36: Add index on `chunks` table for `block_hash`
Confirm that the following query is immediate (it currently takes around 2-4 minutes):

```sql
select * from chunks where block_hash = '51DkcKsnEfPgBwuBCbNTTEfHDR6P8dXWN9r2PfYBXnRE' limit 1;
```

khorolets:
It appears to happen only in case of data-receipts, because we can't find previous receipt by `receipt_id`. Instead we need to look for receipt via `data_id` through `receipts_action_output_data`. I've changed the way we're looking for parent transaction for receipts to differentiate action-receipts and data-receipts.

fixes #39 
also closes #36

### DIFF
--- a/migrations/2020-11-06-103108_data_id_add_index/down.sql
+++ b/migrations/2020-11-06-103108_data_id_add_index/down.sql
@@ -1,0 +1,1 @@
+DROP INDEX action_output_data_id_idx;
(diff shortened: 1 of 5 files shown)


bowenwang1996:
why do you need to store data receipts?

khorolets:
@bowenwang1996 to represent what is happening on the chain in Explorer

frol:
@bowenwang1996 Our goal is to represent all the events happening on the network. Do you think it is pointless to expose data receipts to contract developers?

frol (changes requested):
We need the index over the `data_id` as otherwise, it will be too slow with the number of growing data receipts.

Other than that, everything looks good to me!

frol (approved):

PR #45, opened 2020-11-11
feat: Add access keys from genesis to store genesis handler

### LINKED ISSUE #27: Add access keys from genesis to store genesis handler

khorolets:
Closes #27

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -142,8 +142,11 @@ fn main() {
             if args.store_genesis {
                 let near_config = indexer.near_config().clone();
                 actix::spawn(db_adapters::accounts::store_accounts_from_genesis(
-                    near_config,
+                    near_config.clone(),
                 ));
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -142,8 +142,11 @@ fn main() {
             if args.store_genesis {
                 let near_config = indexer.near_config().clone();
                 actix::spawn(db_adapters::accounts::store_accounts_from_genesis(
-                    near_config,
+                    near_config.clone(),
```
frol, line 145: Can we use a reference to the config? (Don't spend more than 5 minutes on this task, keep the clone if the reference does not work well in this case)

frol (approved):

src/main.rs
khorolets: borrowed values doesn't live long enough and other stuff about borrowing for `'static`

frol (approved):

PR #53, opened 2020-11-23
fix: Investigate missing chunks

### LINKED ISSUE #48: Investigate missing chunks
While syncing testnet from genesis, we've encountered a lot retries for storing receipts, they are infinite because they can't find the parent. Investigation of the reason showed that the parent transactions for these receipts were not stored and then we've found out that the chunk which contains particular parent transactions hasn't been added to database.

For example this one chunk hash `H6MvjzT6JtqzuJiPNGennqBDhtqXfDcb2e8cL5NLNcgU` (block height 22303582)

We need to investigate why Indexer hasn't store this chunks and similar ones and fix it.

/cc @frol

khorolets:
Fixes #48 

As we investigated we saw that nothing is missing actually but sometimes it is delayed to be stores because of collisions between retries. We decided to increase the delay between attempts up to 2 min to give them a chance to be applied successfully.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1914,7 +1914,7 @@ dependencies = [
 [[package]]
 name = "near-actix-utils"
 version = "0.1.0"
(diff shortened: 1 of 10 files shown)


frol (approved):

frol:
P.S. 10000ms is 10 seconds, not 2 minutes

frol (approved):

PR #55, opened 2020-11-27
refactor: Use buffer_unordered to allow only 100 spawned processes at a time to decrease collision

khorolets:


frol (approved):
That was too easy ;-)

PR #63, opened 2021-01-14
fix: Fill delete_by_receipt_id of access keys when owner account is deleted

### LINKED ISSUE #61: bug: deleted_by_receipt_id for access_keys of deleted account should not be null
@vgrichina asked in Discord
> I wonder what is the expected behavior of indexer when account is deleted?
>
> I've just tried creating test-indexer-keys.vg with single key (ed25519:FUMSeS84AsKAq8ydAWiE5io1CxC9NSdAW74EGPyxmKPL).
>
> Then I deleted the account, however deleted_by_receipit_id is still null. Is this by design?
>
> I wonder if there is any way to determine from indexer DB whether access key is still active.

I have clarified the flow with @frol and we agreed that Indexer should set `deleted_by_receipt_id` for access keys of the deleting account with receipt id which deleted the account.

It will require additional re-index from genesis to fullfill the field for missing access keys (on mainnet it will take ~3 days to complete)

khorolets:
Fixes #61

### DIFF
--- a/src/db_adapters/access_keys.rs
+++ b/src/db_adapters/access_keys.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::convert::TryFrom;
 
+use bigdecimal::BigDecimal;
 use diesel::r2d2::{ConnectionManager, Pool};
 use diesel::{ExpressionMethods, PgConnection, QueryDsl};
 use futures::{join, StreamExt};
@@ -33,11 +34,24 @@ pub(crate) async fn handle_access_keys(
         .filter_map(|outcome_with_receipt| outcome_with_receipt.receipt.as_ref());
 
     let mut access_keys = HashMap::<(String, String), models::access_keys::AccessKey>::new();
+    let mut deleted_accounts = HashMap::<String, String>::new();
 
     for receipt in successful_receipts {
         if let near_primitives::views::ReceiptEnumView::Action { actions, .. } = &receipt.receipt {
             for action in actions {
                 match action {
+                    near_primitives::views::ActionView::DeleteAccount { .. } => {
+                        deleted_accounts.insert(
+                            receipt.receiver_id.to_string(),
+                            receipt.receipt_id.to_string(),
+                        );
+                        access_keys.iter_mut().for_each(|(key, value)| {
+                            if key.1 == receipt.receiver_id.to_string() {
+                                value.deleted_by_receipt_id = Some(receipt.receipt_id.to_string());
+                                value.last_update_block_height = block_height.into();
+                            }
+                        });
+                    }
                     near_primitives::views::ActionView::AddKey {
                         public_key,
                         access_key,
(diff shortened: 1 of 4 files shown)


src/db_adapters/accounts.rs
```diff
@@ -62,6 +62,35 @@ pub(crate) async fn handle_accounts(
                         }
                     }
                     near_primitives::views::ActionView::DeleteAccount { .. } => {
+                        let target = schema::access_keys::table.filter(
+                            schema::access_keys::dsl::account_id
+                                .eq(receipt.receiver_id.clone().to_string()),
+                        );
```
frol, line 68: 1. This whole handling should belong to the `access_keys` adapter (not only from the structure point of view, but also to avoid data race between `handle_access_keys` and `handle_accounts` which we run in parallel in `handle_message`)
2. This way of updating the keys that belong to the deleted account is not idempotent (as a rule of thumb, consider a usecase when we have a fully synced database and we run another indexer from genesis: if we have all the accounts and access keys in the latest state and then we start from genesis and somewhere in history the account has been deleted, it does not mean that we want to delete the access keys [further in the history the account could be recreated and new keys added]), we should also take `last_update_block_height` into account.

frol (changes requested):

frol:
Consider the following corner cases (the order matters):

* AddKey, DeleteAccount in the same transaction (the key should be recorded and marked as deleted)
* AddKey, DeleteAccount in the same block (the key should be recorded and marked as deleted)
* DeleteAccount, CreateAccount, AddKey in the same block (the key should be recorded and NOT marked as deleted)

Notes:
* DeleteAccount, AddKey is impossible in a single transaction since DeleteAccount can only be the last action
* DeleteAccount, AddKey in the same block cannot happen since AddKey should fail when there is no account

src/db_adapters/access_keys.rs
```diff
@@ -33,11 +34,24 @@ pub(crate) async fn handle_access_keys(
...
                 match action {
+                    near_primitives::views::ActionView::DeleteAccount { .. } => {
+                        deleted_accounts.insert(
+                            receipt.receiver_id.to_string(),
+                            receipt.receipt_id.to_string(),
+                        );
+                        access_keys.iter_mut().for_each(|(key, value)| {
+                            if key.1 == receipt.receiver_id.to_string() {
+                                value.deleted_by_receipt_id = Some(receipt.receipt_id.to_string());
+                                value.last_update_block_height = block_height.into();
```
frol, line 51: We are collecting the HashMap for a single block, so if the value is in the HashMap, it already has the correct block height provided on initialization, so we can remove this line.

```suggestion
```

src/db_adapters/access_keys.rs
```diff
@@ -33,11 +34,24 @@ pub(crate) async fn handle_access_keys(
...
     for receipt in successful_receipts {
         if let near_primitives::views::ReceiptEnumView::Action { actions, .. } = &receipt.receipt {
             for action in actions {
                 match action {
+                    near_primitives::views::ActionView::DeleteAccount { .. } => {
+                        deleted_accounts.insert(
+                            receipt.receiver_id.to_string(),
+                            receipt.receipt_id.to_string(),
+                        );
+                        access_keys.iter_mut().for_each(|(key, value)| {
```
frol, line 48: To be honest, I find simple `for` loops easier to read than `for_each` in such scenarios

```suggestion
                        for ((_, receiver_id), access_key) in access_keys.iter_mut() {
```

Alternatively, if I use functional notation, I would use filter:

```rust
access_keys.iter_mut()
    .filter(|((_, receiver_id), _)| receiver_id == receipt.receiver_id)
    .for_each(|(_, access_key)| {
         access_key.deleted_by_receipt_id = Some(receipt.receipt_id.to_string());
    })
```

src/db_adapters/access_keys.rs
```diff
@@ -33,11 +34,24 @@ pub(crate) async fn handle_access_keys(
...
+                            if key.1 == receipt.receiver_id.to_string() {
```
frol, line 49: I would expect this to work fine :thinking: 

```suggestion
                            if key.1 == receipt.receiver_id {
```

src/db_adapters/access_keys.rs
```diff
@@ -109,6 +123,47 @@ pub(crate) async fn handle_access_keys(
...
+    let delete_access_keys_for_deleted_accounts = async {
+        let last_update_block_height: BigDecimal = block_height.into();
+        for (account_id, deleted_by_receipt_id) in deleted_accounts {
+            let target = schema::access_keys::table
+                .filter(schema::access_keys::dsl::deleted_by_receipt_id.is_null())
+                .filter(
+                    schema::access_keys::dsl::last_update_block_height
+                        .lt(last_update_block_height.clone()),
+                )
+                .filter(schema::access_keys::dsl::account_id.eq(account_id));
```
frol, line 135: Perfecto!

frol (changes requested):

frol (approved):

PR #69, opened 2021-02-23
feat(buildkite): initial setup

chefsale:
Test job to verify this works: https://buildkite.com/nearprotocol/near-indexer-for-explorer/builds/1. After fail fixed formatting.


mhalambek (approved):

frol:
@chefsale @mhalambek You are awesome! Thank you for integrating CI where we lag behind!

PR #73, opened 2021-03-12
feat: run local using docker

mhalambek:


Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN curl https://sh.rustup.rs -sSf | \
+    sh -s -- -y --no-modify-path --default-toolchain "$(cat /tmp/rust-toolchain)"
+
+RUN cargo install diesel_cli --no-default-features --features "postgres"
+COPY . .
+RUN cargo build --release -p indexer-explorer
```
frol, line 28: I would create a second stage extracting the binary itself, and the nearcore home folder to greatly slim down the image

.buildkite/pipeline.yml
```diff
@@ -1,12 +1,8 @@
...
-    label: "Formatter"
-    
-  - command: "source ~/.cargo/env && cargo clippy"
-    label: "Linter"
-    
-  - command: "source ~/.cargo/env && cargo test"
-    label: "Tests"
-
-  - command: "fossa init && fossa analyze --server-scan && fossa test"
-    label: "Fossa"
```
frol, line 12: Why do you believe it is not needed?

frol (changes requested):

.buildkite/pipeline.yml
chefsale: @mhalambek we cannot delete this.

chefsale: I believe that @mhalambek wanted to add this as a seperate pipeline, but didnt realized there was already this config. Rename the new one to docker-release.yml and leave the original one as is.

chefsale (changes requested):

Dockerfile
mhalambek: Did some experiments with it, it's not as easy, there are some runtime libs it requires, I've tried it using musl but it wasn't working properly

.buildkite/pipeline.yml
mhalambek: I've deleted this by mistake, forgot to check if there is already something there

frol:
@mhalambek do you still want to include the docker image build in the `.buildkite/pipeline.yml`? It is gone now

frol (changes requested):
@mhalambek Just a friendly reminder ^

Dockerfile
khorolets: It needs only `libpq-dev` if I'm not mistaken

frol:
@mhalambek I see a force-push with no changes. Was it intentional?

Issue #74, opened 2021-03-13
Duplicated rows in DB


telezhnaya:
Try executing following query:
```
select * from account_changes 
where account_changes.changed_in_block_hash = 'CFAAJTVsw5y4GmMKNmuTNybxFJtapKcrarsTh5TPUyQf';
```

Expected: 5 rows
Actual: 30 rows, that could be easily grouped to 5

Similar request to RPC:
```

POST to https://archival-rpc.mainnet.near.org
Body
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "method": "EXPERIMENTAL_changes",
  "params": {
    "changes_type": "account_changes",
    "account_ids": ["nfvalidator1.near", "nfvalidator2.near", "nfvalidator3.near", "nfvalidator4.near", "near", "erm.near", "treasury.near"],
     "block_id": 9820214
  }
}
```
Result gives 5 rows

Looks like we have an error in unique constraint, and we do not handle null values properly. See this article for more details:
https://www.enterprisedb.com/postgres-tutorials/postgresql-unique-constraint-null-allowing-only-one-null

rozgo:
Column `id` is required in `account_changes` table, but its not being set, nor auto-incremented.

frol:
@rozgo the column is `bigserial`, so it is auto-incremented. Do you experience any issues on your end?

rozgo:
Many variations of this issue trying to get the indexer working:
```
run --store-genesis --allow-missing-relations-in-first-blocks 1000 sync-from-latest`
Mar 23 18:25:05.898  INFO indexer_for_explorer: Adding/updating accounts from genesis...
Mar 23 18:25:05.898  INFO indexer_for_explorer: Adding/updating access keys from genesis...
Mar 23 18:25:05.900  INFO stats: Server listening at ed25519:...@0.0.0.0:24571
Mar 23 18:25:07.329 ERROR indexer_for_explorer: Error occurred while Accounts from genesis were being added to database. Retrying in 100 milliseconds... 
 Execute(
    DatabaseError(
        __Unknown,
        "null value in column \"id\" violates not-null constraint",
    ),
)
 
```

frol:
Ouch, indeed, I was looking into an old database schema, and now looking into the right migration scripts I see that we use `bigint` instead of `bigserial` there. Could you test the fix: #83? The original issue is not related to this issue with the `id`.

telezhnaya:
[...]

PR #83, opened 2021-03-23
fix: Fixed the type for accounts.id and account_changes.id to be auto-incrementing (bigserial)

frol:
Resolves https://github.com/near/near-indexer-for-explorer/issues/74#issuecomment-800721467

We failed to squash the 0.2.x migrations properly and did not stumble upon the issue before since we gradually upgraded the database instead of wiping it completely on our end.

There is no need for a new migration since with the original migration, nothing could be indexed properly anyway, so we just need to fix that initial migration.


frol:
cc @rozgo

rozgo:
Tested. Accounts work again. Error is gone.

PR #86, opened 2021-04-05
docs: Added public read-only credentials to the Indexer for Explorer database

frol:
NOTE: The database instances referenced in the README are read-only replicas.


telezhnaya:
Works like charm!
All modifying operations break with `ERROR:  cannot execute CREATE ROLE in a read-only transaction`
Postgres gives us guarantee that it's not possible to change anything
https://www.postgresql.org/docs/9.1/sql-set-transaction.html

telezhnaya (approved):

PR #94, opened 2021-04-14
feat: Improve sync-from-interruption mode

### LINKED ISSUE #93: Improve `sync-from-interruption` mode
In order to automate Indexer for Explorer CI/CD we want indexer to be able to decide from which block it should start indexing. We already have `sync-from-interruption` mode but in our case we want to start indexing from N blocks earlier than indexer was interrupted.

To avoid adding unnecessary features on the NEAR Indexer Framework side it was suggested to handle this on the Indexer for Explorer side.

We want to perform a trick if `sync-from-interruption` mode was passed:
- read explorer database to get the height of the latest block we know (`LATEST`)
- calculate `START_FROM = LATEST - N`
- start indexer by replacing `sync-from-interruption` with `sync-from-block --height START_FROM` under the hood

This will allow to automate indexer restarting in case of any fails.

/cc @chefsale

khorolets:
Resolves #93

### DIFF
--- a/src/db_adapters/blocks.rs
+++ b/src/db_adapters/blocks.rs
@@ -44,21 +44,12 @@ pub(crate) async fn store_block(
 /// Gets the latest block's height from database
 pub(crate) async fn latest_block_height(
     pool: &actix_diesel::Database<PgConnection>,
-) -> Result<u64, String> {
-    let result: Vec<models::blocks::Block> = schema::blocks::table
-        .limit(1)
+) -> Result<Option<u64>, String> {
+    Ok(schema::blocks::table
+        .select((schema::blocks::dsl::block_height,))
         .order(schema::blocks::dsl::block_height.desc())
-        .load_async(&pool)
+        .get_optional_result_async::<(bigdecimal::BigDecimal,)>(&pool)
         .await
-        .unwrap_or_else(|_| vec![]);
-
-    if result.is_empty() {
-        return Err("No blocks found in database".to_string());
-    }
-
-    Ok(result[0]
-        .block_height
-        .clone()
-        .to_u64()
-        .expect("Failed to convert block_height from BigDecimal to u64"))
+        .map_err(|err| format!("DB Error: {}", err))?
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -125,6 +125,77 @@ async fn listen_blocks(
...
+        // without any changes
+        if interruption_args.delta == 0 {
+            return near_indexer::IndexerConfig {
+                home_dir,
+                sync_mode: near_indexer::SyncModeEnum::FromInterruption,
+                await_for_node_synced: if args.stream_while_syncing {
+                    near_indexer::AwaitForNodeSyncedEnum::StreamWhileSyncing
+                } else {
+                    near_indexer::AwaitForNodeSyncedEnum::WaitForFullSync
+                },
```
frol, line 146: Extract this to a variable before the line 133, so you don't need to copy-paste this logic over and over again.

src/db_adapters/blocks.rs
```diff
@@ -39,3 +40,25 @@ pub(crate) async fn store_block(
...
+
+    if result.is_empty() {
+        return Err("No blocks found in database".to_string());
+    }
+
+    Ok(result[0]
+        .block_height
+        .clone()
+        .to_u64()
+        .expect("Failed to convert block_height from BigDecimal to u64"))
```
frol, line 63: With the following things it could be easier to handle (btw, it might be useful to return `Option<u64>` or even `Option<near_primitives::BlockHeight>`

```suggestion
        .select((schema::blocks::columns::BlockHeight))
        .first_async(&pool)
        .await
        .optional()
        .map_err(|err| format!("DB Error: {}", err))?
        .map(|(block_height,)| block_height.to_u64())
```

frol (approved):
Overall, looks good to me!

chefsale:
Thanks a lot for implementing this, I'm finalizing the terraform configuration for spinning up the machines, so this is perfect timing. 🥇

frol (approved):

PR #96, opened 2021-04-21
refactor: Upgrade NEAR Indexer dependency to 0.9.0. 

khorolets:
Make corresponding changes to db_adapters:
* chunks
* transaction
* receipts
* access_keys
* accounts

This is a work in progress PR, we still need to decide what to do with Execution Outcomes as they actually don't belong to specific chunk but to specific shard. 

Perhaps we need to remove the relation `exectuion_outcomes.executed_in_chunk_hash` and add field `shard_id` to be more accurate with data representation we have. Need to discuss and decide, but currently `db_adapters::execution_outcomes` is broken completely.


src/models/mod.rs
```diff
@@ -28,10 +28,13 @@ pub mod receipts;
 mod serializers;
 pub mod transactions;
 
-pub(crate) fn establish_connection() -> actix_diesel::Database<PgConnection> {
+pub(crate) fn get_database_credentials() -> String {
     dotenv().ok();
 
-    let database_url = env::var("DATABASE_URL")
-        .unwrap_or_else(|_| panic!("DATABASE_URL must be set in .env file"));
+    env::var("DATABASE_URL").unwrap_or_else(|_| panic!("DATABASE_URL must be set in .env file"))
```
frol, line 34: This is equivalent to:

```suggestion
    env::var("DATABASE_URL").expect("DATABASE_URL must be set in .env file")
```

src/main.rs
```diff
@@ -207,6 +209,10 @@ fn main() {
     // (sending telemetry and downloading genesis)
     openssl_probe::init_ssl_cert_env_vars();
 
+    // This is a sanity check. Indexer should fail .env file with credentials is missing
+    // We prefer to fail as soon as possible to avoid heavy running for nothing
+    models::get_database_credentials();
```
frol, line 214: FYI, from the call sight it is not immediately obvious why it should fail. We ask to get the credentials and don't check anything, and discard the result completely. I know we panic there, but it is not ideal. Still, it works.

I propose we add the `database_url` to the clap structure and configure it with `DATABASE_URL` env variable: https://docs.rs/structopt/0.3.21/structopt/?search=#environment-variable-fallback, this way we will get a nice error message from clap, and we don't need to have this hacky function that goes around the CLI arguments

frol (approved):

src/main.rs
khorolets: I've extracted it to separate issue #97 to deal with it later.

khorolets:
@chefsale could you look at failing FOSA check to help with identifying the nature of the error?

frol (approved):
LGTM. I have no clue where even to look into FOSSA report :disappointed: (CI does not give any hints and FOSSA web-app does not show this PR or at least I could not find it)

PR #99, opened 2021-04-28
refactor: Denormalize action_receipt_actions table in order to improve common queries performance

### LINKED ISSUE #87: Denormalize `action_receipt_actions` table in order to improve common queries performance
* [x] add columns to `action_receipt_actions` table and enable indices for them:
  * [x] `predecessor_account_id`
  * [x] `receiver_account_id`
  * [x] `included_in_block_timestamp`
  
  The motivation is that JOINing with `receipts` table while doing filtering like `receiver_account_id = ? AND action_kind = ? AND args = ?` is slow. @vgrichina Could you provide a minimal useful query?
  
  Ideally, we want to have cross-table indices, but Postgres does not support that, though Postgres 12 and 13 claim to improve various aspects of index performance, so we may want to try to migrate our Postgres 11 database to Postgres 13 to see if that helps.
* [x] add index on `args->>'method_name'` for `FUNCTION_CALL` actions (conditional index in Postgres)
* [x] try to decode base64-encoded `args->>args_base64` to JSON (replace the original base64 field with `args->>args_json` if parsing succeeds), so ultimately we will either have `args_base64` (e.g. rainbow bridge uses borsh instead of JSON args) or `args_json` (commonly used by NEAR smart contracts)
  * [x] consider adding indices to `args->>'args_json'->>'receiver_id'` and `args->>'args_json'->>'amount'` for FT and NFT

khorolets:
Closes #87 

Starting to address the issue. Plan:

- [x] Migrations to add columns and fill them with the data
- [x] Update corresponding model (`ActionReceiptAction`)
- [x] Update corresponding db adapter which stores ActionReceipt actions to include data for newly added columns
- [x] Adjust adapter to extend `args` field to include `args_base64` (is there from the start) and `args_json` if possible (in case of successful deserialization)
- [x] (considerable) Add indexes on `args->>'args_json'->>'receiver_id'` and `args->>'args_json'->>'amount'`
- [x] create migration script/command to iterate over all function call args and try to parse `args_json` to fill historical data along with newer data


src/db_adapters/receipts.rs
```diff
@@ -362,6 +363,9 @@ async fn store_receipt_actions(
                         receipt.receipt_id.to_string(),
                         i32::from_usize(index).expect("We expect usize to not overflow i32 here"),
                         action,
+                        receipt.clone().predecessor_id.to_string(),
+                        receipt.clone().receiver_id.to_string(),
```
frol, line 367: Why do you clone the whole receipt? I believe you don't need clone here at all

khorolets:
@frol please review, I've checked args logic on localnet and it works. I am hesitating about the new version. I've put `0.6.1` as I don't think we've changed a lot to bump to `0.7.x` any objections?

migrations/2021-04-28-154439_denormalize_action_receipt_actions/up.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE action_receipt_actions
+    ADD COLUMN predecessor_account_id text NOT NULL DEFAULT '',
+    ADD COLUMN receiver_account_id text NOT NULL DEFAULT '',
+    ADD COLUMN included_in_block_timestamp numeric(20, 0) NOT NULL DEFAULT 0;
```
frol, line 4: We need indices for these columns

frol (approved):
LGTM

migrations/2021-04-28-154439_denormalize_action_receipt_actions/up.sql
vgrichina: It's also very desired to have:
- index on `args->>'method_name'`
- `base64_args` decoded into JSON field when possible, so that we can add indices on stuff like `receiver_id` for custom token transfers, etc

vgrichina: oh wait, looks like it's already there 🤦 🤣

migrations/2021-04-28-154439_denormalize_action_receipt_actions/up.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE action_receipt_actions
+    ALTER COLUMN predecessor_account_id DROP DEFAULT,
+    ALTER COLUMN receiver_account_id DROP DEFAULT,
+    ALTER COLUMN included_in_block_timestamp DROP DEFAULT;
+
+CREATE INDEX action_receipt_actions_args_function_call_idx ON action_receipt_actions((args->>'method_name'))
+    WHERE action_receipt_actions.action_kind = 'FUNCTION_CALL';
+
+CREATE INDEX action_receipt_actions_args_amount_idx ON action_receipt_actions((args->'args_json'->>'amount'))
+    WHERE action_receipt_actions.action_kind = 'FUNCTION_CALL' AND (action_receipt_actions.args->>'args_json') IS NOT NULL;
```
frol, line 24: Reminder: we need to migrate the existing args to have args_json

khorolets: I don't like the idea to make such migration on the SQL side. Do you mind if I create additional command for that? I realise it will hung with us there forever, but I prefer to make such things via Rust rather that SQL.

frol: I’m fine with either SQL or Rust script

vgrichina: @khorolets hmm, not sure why it's an issue  Postgres has no problem parsing base64, e.g.:

```sql
select * from (
    select date,
         -- TODO: How to filter only successful?
        count(distinct args->>'receiver_id') as receivers,
        sum((args->>'amount')::numeric) / 1e24 as volume
    from (
        select receipt_id,
            predecessor_account_id,
            receiver_account_id,
            args->>'method_name' as method_name,
            convert_from(decode(args->>'args_base64', 'base64'), 'UTF8')::json as args,
            date_trunc('week', to_timestamp(included_in_block_timestamp / 1000000000)) as date
        from action_receipt_actions join receipts using(receipt_id)
        where predecessor_account_id = :contract_name and receiver_account_id = 'wrap.near'
    ) as with_fields
    where method_name = 'ft_transfer'
    group by date
) as call_stats
order by date asc limit 100
```

so to migrate args via SQL you just need something like

```sql
UPDATE action_receipt_actions
SET args_base64 =  convert_from(decode(args->>'args_base64', 'base64'), 'UTF8')::json
```

which you can add to existing `UPDATE ... FROM` statement you have for other columns – that should be the fastest way to pull it off (loading to Rust and pushing back likely gonna take forever).

khorolets: @vgrichina thanks, I've tried but the migration which tried to decode failed on the records where `args_base64` was actual borsh args and it stopped migration. Do you have any suggestions how we can ignore those failing records and don't interrupt the process?

vgrichina: @khorolets yeah, looks like it's a bit tricky but you can handle errors in Postgres:

```plpgsql
CREATE OR REPLACE FUNCTION decode_or_null(bytea) RETURNS json
   LANGUAGE plpgsql AS
$$BEGIN
   RETURN convert_from($1, 'UTF8')::json;
EXCEPTION
   WHEN others THEN
      RAISE WARNING '%', SQLERRM;
      RETURN NULL;
   
END;$$;

select decode_or_null(decode(args->>'args_base64', 'base64')) as args from action_receipt_actions limit 1000;
```

basically you can have your own wrapper function which returns NULL on every exception

khorolets:
For history: While this PR was in progress we've released a few minor version so we end up with `0.6.3` version of Indexer for Explorer in this PR.

frol (approved):

khorolets:
Out data is too big and transaction-based migration queries are locking tables for huge amount of time. We end up in using custom python script splitting big query into smaller ones. Here's the script for the future use.

```python
from sqlalchemy import create_engine

START = 1595370903490523743
STEP = 1000000000000
END = 1620717648739546794
ESTIMATED_STEPS = (END - START) / STEP
connection_string = 'postgresql+psycopg2://user:pass@host/database'

def generate_sql(from_timestamp: int, to_timestamp: int) -> str:
    """
    Generates str for SQL query to convert args_base64 to args_json if possible
    """
    return """
    UPDATE action_receipt_actions
            SET args = jsonb_set(args, '{args_json}', decode_or_null(decode(args->>'args_base64', 'base64')), true)
            WHERE action_kind = 'FUNCTION_CALL'
                AND receipt_receiver_account_id != 'client.bridge.near'
                AND receipt_included_in_block_timestamp <= %d
                AND receipt_included_in_block_timestamp >= %d;
    """ % (to_timestamp, from_timestamp)

if __name__ == '__main__':
    print("Establishing connection to %s..." % (connection_string.split('/')[-1],))
    engine = create_engine(connection_string)
    print(f"Estimated queries to execute: {ESTIMATED_STEPS}.")
    from_timestamp = START-STEP
    to_timestamp = START
    counter = 1
    with engine.connect() as con:
        while True:
            from_timestamp += STEP
            to_timestamp += STEP
            if (END - to_timestamp) < STEP:
                break
            print(f"{counter}/{ESTIMATED_STEPS} (from {from_timestamp} to {to_timestamp}")

            out = con.execute(generate_sql(from_timestamp, to_timestamp))
            print(out.rowcount)
            counter += 1

    print("FINISHED")

```

PR #107, opened 2021-05-17
fix: Handle recreation of implicit accounts properly to avoid confusion for the users

### LINKED ISSUE #68: Keep the `created_by_receipt_id` for implicit account pointing to the first TRANSFER action rather updating it on every TRANSFER action
@khorolets we had this conversation before, but it seems that the current behavior confuses end-users (https://github.com/near/near-explorer/issues/551). Updating the creation receipt id for implicit accounts causes confusion as to why there are transactions that are before the "creation" transaction. In the common case, people would rather expect to see that the very first transaction causes the creation; in our case we should make sure that we handle the account deletion properly, but we should be able to only update the creation receipt id when deletion receipt id is not NULL (in which case we want to set it to NULL and update the creation receipt id).

khorolets:
Fixes #68 

Also I've changed `nearcore` dependency hash to the release hash of [1.19.0](https://github.com/near/nearcore/releases/tag/1.19.0) Only a version has changed in that commit in compare with previous, but I decided to stick to the release as it is perfect timing for that.

I've tried to leave comments and renamed some variables as I've found them a bit misleading.

### DIFF
--- a/src/db_adapters/accounts.rs
+++ b/src/db_adapters/accounts.rs
@@ -80,7 +80,7 @@ pub(crate) async fn handle_accounts(
         }
     }
 
-    let (accounts_to_insert, accounts_to_update): (
+    let (accounts_to_create_or_update, accounts_to_delete): (
         Vec<models::accounts::Account>,
         Vec<models::accounts::Account>,
     ) = accounts
@@ -88,8 +88,8 @@ pub(crate) async fn handle_accounts(
         .cloned()
         .partition(|model| model.created_by_receipt_id.is_some());
 
-    let update_accounts_future = async {
-        for value in accounts_to_update {
+    let delete_accounts_future = async {
+        for value in accounts_to_delete {
             let target = schema::accounts::table
                 .filter(schema::accounts::dsl::account_id.eq(value.account_id))
                 .filter(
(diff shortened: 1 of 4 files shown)


src/db_adapters/accounts.rs
```diff
@@ -152,7 +152,59 @@ pub(crate) async fn handle_accounts(
...
+            Vec<models::accounts::Account>,
+            Vec<models::accounts::Account>,
+        ) = accounts_to_create_or_update.into_iter().partition(|model| {
+            model.account_id.len() == 64 && model.deleted_by_receipt_id.is_none()
+        });
+
+        for value in implicit_accounts_to_recreate {
+            let target = schema::accounts::table
+                .filter(schema::accounts::dsl::account_id.eq(value.account_id))
+                .filter(schema::accounts::dsl::deleted_by_receipt_id.is_not_null()) // this filter ensures we update only "deleted" accounts
```
frol, line 171: There is a rare corner case that I don't think we need to address right now since I don't see a use-case for such case, but I just want to record it:

The database state: existing non-deleted implicit account (deleted_by_receipt_id is NULL)

The receipt: `DeleteAccount` + `Transfer` (re-creating the account immediately)

Expected result: created_by_receipt_id is updated, deleted_by_receipt_id remains NULL

This implementation will lead to completely ignoring this event.

frol (approved):
This implementation looks great to me! Thanks for clarifying the variable names and providing useful references in the comments! :rocket: 

We also need to migrate the database somehow. I think we can reset `last_update_block_height` to 0 and re-indexer from genesis. What do you think?

src/db_adapters/accounts.rs
khorolets: I think this case will be ignored even earlier, where we collect all the accounts here https://github.com/near/near-indexer-for-explorer/blob/b0dbbe52f7bebf3904365805e4a3ccd90ae12c53/src/db_adapters/accounts.rs#L40-L77

Should we merge it as is?

frol: Yes, let's merge as is

Issue #111, opened 2021-05-22
Extend StateChangeReasonKind with new reason Migration


khorolets:
This changes on nearcore https://github.com/near/nearcore/commit/473e71b5127e3795719c793ab727ef629de53d33 have introduced a new variant to the enum which we are not handling. We need to add it along with corresponding migrations into database schema and models.

khorolets:
@telezhnaya while you are around this model/adapter could you address this issue as well, please?

telezhnaya:
@khorolets sure
Is it urgent? 
I want to do something with #77 #74 #84 and then will make this one, OK?

khorolets:
It's not urgent, it's just easy and perfect for "onboarding" :)

Issue #84, opened 2021-03-25
Some rows are missing because of transaction hash collisions


telezhnaya:
In `accounts` table, we have implicit account that still should exist `ae384ca1c2c93a4029e34fb52a5f72adebc5ff10b8b35fdd628ab2e7c874d088`

But we do not have it in RPC, here is the request body
```
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "method": "query",
  "params": {
    "request_type": "view_account",
    "block_id": 33039470,
    "account_id": "ae384ca1c2c93a4029e34fb52a5f72adebc5ff10b8b35fdd628ab2e7c874d088"
  }
}
```

khorolets:
@telezhnaya good catch! Thanks

@frol looks weird. https://explorer.near.org/accounts/ae384ca1c2c93a4029e34fb52a5f72adebc5ff10b8b35fdd628ab2e7c874d088

It looks like accounts was deleted before `create_account` was created on behalf of this account.

telezhnaya:
get another one
`85fe743120593424437b34b05e45af39f2c3de67b5701950d2fd43be1671e544`

frol:
`ae384ca1c2c93a4029e34fb52a5f72adebc5ff10b8b35fdd628ab2e7c874d088` is indeed a problematic case, I will mention you in another issue where you will be able to read more details.

`nflfun.near` and `85fe743120593424437b34b05e45af39f2c3de67b5701950d2fd43be1671e544` look fine to me:

```
http post https://archival-rpc.mainnet.near.org jsonrpc=2.0 id=dontcare method=query \
                                 params:='{
                               "request_type": "view_account",
                               "finality": "final",
                               "account_id": "nflfun.near"
                             }'
{
    "id": "dontcare",
    "jsonrpc": "2.0",
    "result": {
        "amount": "997673579100734400000000",
        "block_hash": "4N4unU4VxZqJ63WAe1BpAYj1haAtGyVc3AZLDvjqBRmC",
        "block_height": 33072649,
        "code_hash": "11111111111111111111111111111111",
        "locked": "0",
        "storage_paid_at": 0,
        "storage_usage": 390
    }
}
```

telezhnaya:
@frol yeap, you are right, I added them by mistake. They were just created later than I needed. 
So we only need to think about `ae384ca1c2c93a4029e34fb52a5f72adebc5ff10b8b35fdd628ab2e7c874d088`

[...]

Issue #77, opened 2021-03-18
Do we miss some balance changes?


frol:
I sum up non-staking and staking-balances based on the latest state changes per account:

```sql
SELECT
    SUM(affected_account_nonstaked_balance),
    SUM(affected_account_staked_balance)
FROM (
    SELECT DISTINCT ON (affected_account_id) affected_account_nonstaked_balance, affected_account_staked_balance
    FROM account_changes
    ORDER BY affected_account_id, changed_in_block_timestamp DESC
) as t;
```

And here are the values:

```
                sum                |                sum
-----------------------------------+-----------------------------------
 617515305435595972472724261340813 | 402992430818950819988234479163231
(1 row)
```

The problem here is that this exceeds the total supply. Note, there are three accounts created in genesis (`bo.near`, `registrar`, `wallet.pulse.near`), which have never been touched, so they are not in the `account_changes` table (they hold just 60 NEAR)

```
>>> total_supply
1019905212545170492542259392171734

>>> 617515305435595972472724261340813 + 402992430818950819988234479163231
1020507736254546792460958740504044

>>> 617515305435595972472724261340813 + 402992430818950819988234479163231 + (50000000000000000000000000 + 10000000000000000000000000 + 999899913398562500000000)
1020507797254446705859521240504044
```

The difference is 602.5k NEAR

Any ideas? @telezhnaya @bowenwang1996

bowenwang1996:
@frol some questions:
* Is this computation done on the same block for which you queried total supply?
* Is it possible that account change implementation is incorrect? For example, are burnt balances accounted for?

frol:
@bowenwang1996

> Is this computation done on the same block for which you queried total supply?

Total supply did not change while the query was computed, I kept querying the `block` JSON RPC endpoint.

> Is it possible that account change implementation is incorrect? For example, are burnt balances accounted for?

It gets the snapshot of the `AccountView` from the state changes and just records it into the database as is (maps `amount` -> `affected_account_nonstaked_balance` and `locked` -> `affected_account_staked_balance`)

The query essentially pulls the latest available state change about each individual account and sums up those latest balances.

bowenwang1996:
> It gets the snapshot of the AccountView from the state changes and just records it into the database as is (maps amount -> affected_account_nonstaked_balance and locked -> affected_account_staked_balance)

[...]

PR #118, opened 2021-05-28
fix: add unique indexes on account_changes table

telezhnaya:
Fixes https://github.com/near/near-indexer-for-explorer/issues/74

### DIFF
--- a/migrations/2021-05-27-154211_account_changes_unique_idx/down.sql
+++ b/migrations/2021-05-27-154211_account_changes_unique_idx/down.sql
@@ -1,0 +1,3 @@
+DROP INDEX IF EXISTS account_changes_transaction_uni_idx;
+DROP INDEX IF EXISTS account_changes_receipt_uni_idx;
+DROP INDEX IF EXISTS account_changes_null_uni_idx;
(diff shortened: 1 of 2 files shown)


khorolets (approved):
Great finding and nice fix!

> It requires full re-deploy of DB
@telezhnaya Why do you say that?

telezhnaya:
> > It requires full re-deploy of DB

> @telezhnaya Why do you say that?

I could use weird wording, sorry, I meant that we need re-fill it, you know about that. We can't put unique index since the table has duplicates

khorolets:
@telezhnaya ah ok, I thought you have seen something else which will require something more stressful. Just for clarification, could you please add refs to issues you want to address before upgrade? Thanks. Good job!

telezhnaya:
https://github.com/near/near-indexer-for-explorer/issues/111
https://github.com/near/near-indexer-for-explorer/issues/84
https://github.com/near/near-indexer-for-explorer/issues/77

Some of them could not require re-filling, but anyway, it's better to be sure in that
I will ping you directly when it would be the time 🙂

frol (approved):
Very well done!

frol:
@telezhnaya Can we somehow remove the duplicates without re-indexing the whole history from scratch?

telezhnaya:
@frol I am afraid we will not be happy with that.
There are 105M lines in this table. I am not highload specialist, but it sounds like suffering pain.

Another reason (the main one, actually. It's OK to suffer) is that if I delete something from this table, I will not trust to that data anymore. Each time when I will not find needed lines, I will think that it's not because there is an error; it's because I am stupid and deleted more than I wanted.

To sum up, let's just wait with this migration and re-collect the data when we anyway will have to re-collect it (or when we forget how hard it was during last time)

PR #120, opened 2021-06-01
fix: unique indexes in account_changes table should contain update_reason

telezhnaya:
See an example:
```
select * from account_changes 
where account_changes.affected_account_id = '0520974a09c1e419a208b9238363df42df30ff94.lockup.near' 
and account_changes.changed_in_block_timestamp = 1604431876987521027 
and account_changes.caused_by_receipt_id = 'HUEGvT8ceLGWgGyx2o8vsSgoRnSEkpgJsi4UbeiaMsXJ'
order by account_changes.changed_in_block_timestamp desc, id desc;
```

In the result, all the columns are the same except `update_reason`: it's `ACTION_RECEIPT_GAS_REWARD` and `RECEIPT_PROCESSING`, in the same block and receipt
So we should put `update_reason` to index also.

This one is also not OK for the same reason, and I'm not sure what should we do with it. Honestly, we could just delete it, it does nothing. Is it OK to remove this line in this PR also?
https://github.com/near/near-indexer-for-explorer/blob/master/migrations/2020-12-07-153402_initial_schema/up.sql#L285

### DIFF
--- a/migrations/2021-05-27-154211_account_changes_unique_idx/up.sql
+++ b/migrations/2021-05-27-154211_account_changes_unique_idx/up.sql
@@ -1,11 +1,11 @@
 CREATE UNIQUE INDEX account_changes_transaction_uni_idx
-ON account_changes (affected_account_id, changed_in_block_hash, caused_by_transaction_hash)
+ON account_changes (affected_account_id, changed_in_block_hash, caused_by_transaction_hash, update_reason)
 WHERE caused_by_transaction_hash IS NOT NULL AND caused_by_receipt_id IS NULL;
 
 CREATE UNIQUE INDEX account_changes_receipt_uni_idx
-ON account_changes (affected_account_id, changed_in_block_hash, caused_by_receipt_id)
+ON account_changes (affected_account_id, changed_in_block_hash, caused_by_receipt_id, update_reason)
 WHERE caused_by_transaction_hash IS NULL AND caused_by_receipt_id IS NOT NULL;
 
 CREATE UNIQUE INDEX account_changes_null_uni_idx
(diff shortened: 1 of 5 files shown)


frol (approved):

frol:
> This one is also not OK for the same reason, and I'm not sure what should we do with it. Honestly, we could just delete it, it does nothing. Is it OK to remove this line in this PR also? https://github.com/near/near-indexer-for-explorer/blob/master/migrations/2020-12-07-153402_initial_schema/up.sql#L285

Let's drop it in the `up.sql` migration we currently update.

frol (approved):
@telezhnaya Please, coordinate the release and re-index with @khorolets

migrations/2021-05-27-154211_account_changes_unique_idx/up.sql
```diff
@@ -1,11 +1,40 @@
+-- This constraint is bad because of 2 reasons:
+-- 1. It does not contain update_reason, so it can remove needed lines
+-- 2. Fortunately, it does not work at all because we have 2 nullable columns, they could not be compared by equality
+--    (all nulls are considered equal)
```
khorolets, line 4: I thought all nulls are considered **unique** and that caused the problem, no?

khorolets (approved):

migrations/2021-05-27-154211_account_changes_unique_idx/up.sql
telezhnaya: Yeap, thanks, that's a typo

khorolets:
@telezhnaya will you bump the version and add some info to changelog? :)

khorolets (approved):

PR #121, opened 2021-06-02
feat: add new change reason kind MIGRATION

### LINKED ISSUE #111: Extend StateChangeReasonKind with new reason Migration

telezhnaya:
Fixes #111

### DIFF
--- a/migrations/2021-06-02-173100_add_migration_state_change_reason_kind/down.sql
+++ b/migrations/2021-06-02-173100_add_migration_state_change_reason_kind/down.sql
@@ -1,0 +1,2 @@
+-- This file is empty since we can't remove items from enum in Postgres.
+-- To be honest, it does not sound as a big problem. `IF_NOT_EXISTS` will prevent re-adding it.
--- a/migrations/2021-06-02-173100_add_migration_state_change_reason_kind/up.sql
+++ b/migrations/2021-06-02-173100_add_migration_state_change_reason_kind/up.sql
@@ -1,0 +1,1 @@
+ALTER TYPE state_change_reason_kind ADD VALUE IF NOT EXISTS 'MIGRATION';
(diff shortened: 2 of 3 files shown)


khorolets (changes requested):
It's great, but also we need to extend corresponding enum and adjust it's impl :)

https://github.com/near/near-indexer-for-explorer/blob/master/src/models/enums.rs#L103-L129

Also I expect tests to fail after you made this change, because it's in newer commit than release one, so you can take latest master commit for now, we will include it in later release

telezhnaya:
> It's great, but also we need to extend corresponding enum and adjust it's impl :)

I am sorry for that 😞 I forgot to search for ids in camelCase in the project.

> Also I expect tests to fail after you made this change, because it's in newer commit than release one, so you can take latest master commit for now, we will include it in later release

Don't worry about that. If we merge this one earlier, I will rebase the release commit, so the order will be OK.

khorolets:
>Don't worry about that. If we merge this one earlier, I will rebase the release commit, so the order will be OK.

We will postpone merging, will create a release based on #120 and after that will merge this, because this is a preparation to `1.20.0` nearcore release.

khorolets (approved):
Nice and easy! Thank you!

khorolets:
@telezhnaya could you please run `cargo fmt`?

telezhnaya:
> could you please run cargo fmt?

@khorolets  It finished silently with no editions, code is clean 👌

khorolets:
>  It finished silently with no editions, code is clean 👌

my bad, I've misread the build error, sorry :)

frol (approved):

PR #122, opened 2021-06-09
Alter default privileges on select on new tables

telezhnaya:
There was an issue that regular users couldn't select from the new table, fixed that.
It's already applied in production, so no action is required.
With this change, all new tables will be available for reading for all readonly users by default.

I guess we need to merge it after https://github.com/near/near-indexer-for-explorer/pull/121 not to be confused by the order of migrations


khorolets (approved):
Great! Love the comments you are leaving, great job!

frol (approved):

PR #136, opened 2021-07-28
feat: compute circulating supply and store it in DB

telezhnaya:
You can take a look at how it works at `34.141.101.9` (ask @khorolets for access)

```
tmux attach
cargo run --release -- --home-dir ~/.near/mainnet run --store-genesis --stream-while-syncing --allow-missing-relations-in-first-blocks 1000 --concurrency 0 sync-from-latest
```

The testing DB is available at `35.233.39.106`, there are some numbers calculated.

It works well in general, but I am still thinking about working with threads. I am killing the thread if anything bad happens. On the one hand, that's exactly what I wanted to achieve. On the other, it's better to try to recreate the thread from time to time. Want to know your opinion, I guess the best solution will be born after the discussion.


telezhnaya:
I fixed formatting, `cargo fmt` is calm both on my machine and at Bohdan's server, but CI formatter is still angry at me.

src/main.rs
```diff
@@ -264,6 +268,9 @@ fn main() {
                     args.concurrency,
                     args.allow_missing_relations_in_first_blocks,
                 ));
+
+                let view_client = indexer.client_actors().0;
```
khorolets, line 272: I propose we add comments before `listen_blocks` spawn that we're spawning regular indexer job and before this spawn that we're running cirulating supply computation (just to help us to distinguish stuff we're running)

src/main.rs
```diff
@@ -1,16 +1,20 @@
...
 #[macro_use]
 extern crate diesel;
+
 use diesel::PgConnection;
 use futures::{join, StreamExt};
 use tokio::sync::mpsc;
 use tracing::info;
 use tracing_subscriber::EnvFilter;
 
+use crate::circulating_supply::circulating_supply_provider::compute_circulating_supply;
```
khorolets, line 14:
```suggestion
use crate::circulating_supply::circulating_supply_provider;
```

src/main.rs
```diff
@@ -264,6 +268,9 @@ fn main() {
                     args.concurrency,
                     args.allow_missing_relations_in_first_blocks,
                 ));
+
+                let view_client = indexer.client_actors().0;
+                actix::spawn(compute_circulating_supply(view_client));
```
khorolets, line 273: I would like to understand this function is not from the same file from the first sight. 

```suggestion
                actix::spawn(circulating_supply_provider::compute_circulating_supply(view_client));
```

src/db_adapters/blocks.rs
```diff
@@ -53,3 +53,16 @@ pub(crate) async fn latest_block_height(
         .map_err(|err| format!("DB Error: {}", err))?
         .and_then(|(block_height,)| block_height.to_u64()))
 }
+
+/// Gets the latest block before given timestamp (nanos) from database
+pub(crate) async fn closest_block_for(
```
khorolets, line 58: While in Rust we can't use named arguments in function call like `closest_block_for(timestamp = 1)` I suggest to add `timestamp` to the function name
```suggestion
pub(crate) async fn closest_block_for_timestamp(
```

src/db_adapters/accounts.rs
```diff
@@ -326,3 +327,44 @@ pub(crate) async fn store_accounts_from_genesis(near_config: near_indexer::NearC
         "Accounts from genesis were added/updated successful."
     );
 }
+
+pub(crate) async fn collect_lockups_for_block(
```
khorolets, line 331:
```suggestion
pub(crate) async fn collect_lockups_for_block_height(
```

I love your habit to comment such cases! Amazing!

src/circulating_supply/user_balance.rs
```diff
@@ -0,0 +1,37 @@
...
+use near_indexer::near_primitives::views::{QueryRequest, QueryResponseKind};
+
+pub async fn get_user_balance(
+    view_client: Addr<ViewClientActor>,
+    account_id: &str,
+    block_height: u64,
+) -> Balance {
+    let block_reference = BlockReference::BlockId(BlockId::Height(block_height));
+    let request = QueryRequest::ViewAccount {
+        account_id: account_id.parse().unwrap(),
```
khorolets, line 13: I don't insist on the error message, you may change it to be more helpful

```suggestion
        account_id: account_id.parse().expect("Failed to parse `account_id`"),
```

src/circulating_supply/user_balance.rs
```diff
@@ -0,0 +1,37 @@
...
+    let account = match account_response.kind {
+        QueryResponseKind::ViewAccount(x) => x,
+        _ => {
+            panic!(
+                "ViewAccount result expected for {}, block {}",
+                account_id, block_height
+            )
+        }
+    };
+    return account.amount;
```
khorolets, line 36:
```suggestion
    account.amount
```

src/circulating_supply/lockup.rs
```diff
@@ -0,0 +1,209 @@
...
+    // `check_transfers_vote` contract method we won't be able to
+    // get proper information based on timestamp, that's why we inject
+    // the `transfer_timestamp` which is phase2 timestamp
+    state.lockup_information.transfers_information = TransfersInformation::TransfersEnabled {
+        transfers_timestamp: U64(TRANSFERS_ENABLED),
+    };
+    return state;
+}
+
+pub async fn get_code_version(
```
khorolets, line 64:
```suggestion
pub(crate) async fn get_code_version(
```

src/circulating_supply/lockup.rs
```diff
@@ -0,0 +1,209 @@
...
+use actix::Addr;
+use near_client::{Query, ViewClientActor};
+use near_indexer::near_primitives::types::{BlockId, BlockReference};
+use near_indexer::near_primitives::views::{QueryRequest, QueryResponseKind};
+use near_sdk::borsh::BorshDeserialize;
+use near_sdk::json_types::{U128, U64};
+
+pub const TRANSFERS_ENABLED: u64 = 1602614338293769340;
+
+pub async fn get_account_state(
```
khorolets, line 13:
```suggestion
pub(crate) async fn get_account_state(
```

src/circulating_supply/lockup.rs
```diff
@@ -0,0 +1,209 @@
...
+    LockupContract, TransfersInformation, VestingInformation, VestingSchedule, WrappedBalance, U256,
+};
+use actix::Addr;
+use near_client::{Query, ViewClientActor};
+use near_indexer::near_primitives::types::{BlockId, BlockReference};
+use near_indexer::near_primitives::views::{QueryRequest, QueryResponseKind};
+use near_sdk::borsh::BorshDeserialize;
+use near_sdk::json_types::{U128, U64};
+
+pub const TRANSFERS_ENABLED: u64 = 1602614338293769340;
```
khorolets, line 11: Put a comment please explaining why this number

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    add_circulating_supply(&pool, &stats).await;
+    info!(
+        target: crate::INDEXER_FOR_EXPLORER,
+        "Circulating supply for the timestamp {} is {}", block_timestamp, circulating_supply
+    );
+}
+
+async fn find_last_yesterday_block_for(
```
khorolets, line 139: Consider replace "yesterday" to be more clear :)

```suggestion
async fn find_last_yesterday_block_for_timestamp(
```

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+fn get_now_timestamp() -> u64 {
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_nanos() as u64
+}
+
+async fn compute_circulating_supply_for(
```
khorolets, line 48:
```suggestion
async fn compute_circulating_supply_for_timestamp(
```

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+use actix::Addr;
+use actix_web::rt::time;
+use std::ops::Div;
+use std::str::FromStr;
+use std::time::SystemTime;
+use tracing::info;
+
+const DAY: std::time::Duration = std::time::Duration::from_secs(86400);
+
+pub(crate) async fn compute_circulating_supply(view_client: Addr<near_client::ViewClientActor>) {
```
khorolets, line 25: Could you please add docstring describing how is it done (mention that it calculates and calls `add_circulating_supply()` that stores data into DB. If I understood correct.

khorolets (approved):
Nice work! A few comments here and there. Overall great effort!

migrations/2021-06-15-151515_circulating_supply_table/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE circulating_supply
```
frol, line 1: We already have a number of tables in the database, so I suggest we have some prefix in the naming for such precomputed tables. I may suggest `stats__circulating_supply` (please, suggest a better one if you have)

migrations/2021-06-15-151515_circulating_supply_table/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+ALTER TABLE ONLY circulating_supply
+    ADD CONSTRAINT circulating_supply_pkey PRIMARY KEY (block_hash);
+
+CREATE INDEX circulating_supply_timestamp_idx ON circulating_supply USING btree (block_timestamp);
+
+ALTER TABLE ONLY circulating_supply
+    ADD CONSTRAINT circulating_supply_fk FOREIGN KEY (block_hash) REFERENCES blocks (block_hash) ON DELETE CASCADE;
+
+CREATE VIEW lockups AS
```
frol, line 21: Apply the same reasoning for the naming as one I listed above, so we don't confuse people with having `accounts` and `lockups` on the same flat level.

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+use crate::models;
+use crate::models::circulating_supply::CirculatingSupply;
+use crate::models::Block;
+use actix::Addr;
+use actix_web::rt::time;
+use std::ops::Div;
+use std::str::FromStr;
+use std::time::SystemTime;
+use tracing::info;
```
frol, line 21: It would be ideal if we can preserve the code-style around the imports: https://github.com/near/near-indexer-for-explorer/blob/master/CONTRIBUTING.md#imports-ordering-and-grouping

It helps me to review the project structure.

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+
+const DAY: std::time::Duration = std::time::Duration::from_secs(86400);
+
+pub(crate) async fn compute_circulating_supply(view_client: Addr<near_client::ViewClientActor>) {
+    let mut current_day = TRANSFERS_ENABLED;
+    while current_day < get_now_timestamp() {
```
frol, line 27: Instead of converting everything down to `u64`, I suggest we convert it the other way around (to `Duration`), so we don't accidentally compare nonsense

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+        compute_circulating_supply_for(&view_client, current_day).await;
+        current_day += DAY.as_nanos() as u64;
+    }
+
+    // We ping the number twice a day to be sure that we computed it, and to deliver it earlier.
+    // If the value is already in the DB, we will just return almost immediately
+    let mut interval = time::interval(DAY.div(2));
```
frol, line 34: I suggest we introduce a separate const for the interval instead of crafting it ad-hoc.

Also, if we restart the indexer at 11:59pm, we will only get the circulating supply for that day 12 hours later, though we could do that in 2 minutes from that time, so I suggest we have a waiting loop that would check the finalized block timestamp and let it go once we cross the day boundary.

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::time::SystemTime;
+use tracing::info;
+
+const DAY: std::time::Duration = std::time::Duration::from_secs(86400);
+
+pub(crate) async fn compute_circulating_supply(view_client: Addr<near_client::ViewClientActor>) {
+    let mut current_day = TRANSFERS_ENABLED;
+    while current_day < get_now_timestamp() {
+        compute_circulating_supply_for(&view_client, current_day).await;
+        current_day += DAY.as_nanos() as u64;
```
frol, line 29: How long does it take to recompute all the history? I feel that might need to be a separate mode triggered explicitly since we reboot indexers somewhat often and this might be wasteful.

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_nanos() as u64
+}
+
+async fn compute_circulating_supply_for(
+    view_client: &Addr<near_client::ViewClientActor>,
+    request_timestamp: u64,
+) {
+    let pool = models::establish_connection();
```
frol, line 52: Let's establish a database connections pool once for the whole indexer. In the worst case, establish the pool for all the background jobs once

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+            info!(
+                target: crate::INDEXER_FOR_EXPLORER,
+                "Circulating supply for the timestamp {} was already computed: {}",
+                block_timestamp,
+                supply
+            );
+            return;
+        }
+        Err(err) => {
+            panic!("Error {}", err);
```
frol, line 83: Let's not crash the whole indexer if we can retry and recover. Just report a `warn!` and keep running (make the decision yourself whether to rely on the parent logic that will retry in X hours anyway or do an extra layer to retry after some short delay)

src/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    let lockups = collect_lockups_for_block(block_height, &pool).await;
```
frol, line 87: Are they list of lockup account ids? The naming is misleading since I read it as such that the method already returns a list of lockup information structs with all the necessary information about them.

src/circulating_supply/lockup.rs
```diff
@@ -0,0 +1,209 @@
...
+
+pub async fn get_account_state(
+    view_client: &Addr<ViewClientActor>,
+    account_id: &String,
```
frol, line 15: Use `&str`

frol (changes requested):
I have not finished the full review just yet, but decided to share the preliminary review since it is already not tiny.

migrations/2021-06-15-151515_circulating_supply_table/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE circulating_supply
+(
+    block_timestamp          numeric(20, 0) NOT NULL,
+    block_hash               text           NOT NULL,
+    value                    numeric(45, 0) NOT NULL,
+    total_supply             numeric(45, 0) NOT NULL,
+    lockups_number           numeric(45, 0) NOT NULL,
+    active_lockups_number    numeric(45, 0) NOT NULL,
```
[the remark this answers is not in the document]
telezhnaya: thoughts about naming:
total
in_progress
?

migrations/2021-06-15-151515_circulating_supply_table/up.sql
telezhnaya: Thoughts, ideas about prefix:
aggregation
aggregations
views
analytics
stats
statistics
cache
2 prefixes?

Come up with `aggregated__circulating_supply`

src/circulating_supply/circulating_supply_provider.rs
telezhnaya: Todo: put a check here if it was already computed
and delete this check from the algorithm itself

telezhnaya:
is_bug_inside
add the comment what the bug is

src/circulating_supply/circulating_supply_provider.rs
telezhnaya: Yesterday sounds like the most suitable option, sorry 😄 
This is exactly what we want to find there

Tried to play with `day_before`, it always sounds weirder

src/circulating_supply/circulating_supply_provider.rs
telezhnaya: It's better now, but I still clone it sometimes. I tried
```rust
let pool: &'static Database<PgConnection> = &models::establish_connection();
```
It solves the problem with passing the value to any threads, but it gives an error
```
error[E0716]: temporary value dropped while borrowed
   --> src/main.rs:220:50
    |
220 |     let pool: &'static Database<PgConnection> = &models::establish_connection();
    |               -------------------------------    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
    |               |
    |               type annotation requires that borrow lasts for `'static`
...
298 | }
    | - temporary value is freed at the end of this statement
```
Tried to fight with it: no chance for me

migrations/2021-06-15-151515_circulating_supply_table/up.sql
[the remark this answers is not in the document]
telezhnaya: Still not sure about the option I've selected. What do you think?
`unfinished_lockup_contracts_count`

telezhnaya:
Thank you so much @frol @khorolets, you've done amazing amount or work reviewing this.
I've rewritten it almost from scratch 😄 
Had to rebase it to the current master. All new changes are in the second commit.
Please have a look again.

src/aggregated/account_details.rs
```diff
@@ -0,0 +1,64 @@
...
+    view_client: &Addr<ViewClientActor>,
+    account_id: &str,
+    block_height: u64,
+) -> Result<CryptoHash, String> {
+    get_account_view_for_block_height(view_client, account_id, block_height)
+        .await
+        .map(|acc| acc.code_hash)
+}
+
+async fn get_account_view_for_block_height(
```
telezhnaya, line 28: It's enough to have only one call to this function, understood it only now
Will fix it tomorrow

telezhnaya: UPD: I call them for different accounts, so everything is correct. I need balance for foundation accs, and I need code hash for lockups. False alarm 😄

src/db_adapters/accounts.rs
```diff
@@ -326,3 +330,48 @@ pub(crate) async fn store_accounts_from_genesis(near_config: near_indexer::NearC
         "Accounts from genesis were added/updated successful."
     );
 }
+
+pub(crate) async fn get_lockup_ids_at_block_height(
```
frol, line 334:
```suggestion
pub(crate) async fn get_lockup_account_ids_at_block_height(
```

src/db_adapters/accounts.rs
```diff
@@ -326,3 +330,48 @@ pub(crate) async fn store_accounts_from_genesis(near_config: near_indexer::NearC
...
+    pool: &actix_diesel::Database<PgConnection>,
+    block_height: u64,
+) -> Result<Vec<String>, String> {
```
frol, line 337: I suggest returning `Vec<AccountId>`, so the code beyond this place can safely assume that the account_id is valid.

src/aggregated/account_details.rs
```diff
@@ -0,0 +1,64 @@
...
+    view_client: &Addr<ViewClientActor>,
+    account_id: &str,
+    block_height: u64,
+) -> Result<AccountView, String> {
+    let block_reference = BlockReference::BlockId(BlockId::Height(block_height));
+    let request = QueryRequest::ViewAccount {
+        account_id: account_id
+            .parse()
+            .map_err(|_| "Failed to parse `account_id`")?,
```
frol, line 37: I suggest this function requires `account_id` to be `AccountId`, and all the functions above should also require `AccountId`, and validate the account id when we read the value (e.g. when it is a foundation account, we can use `"foundation.near".parse().expect()` since those are obviously valid).

src/aggregated/account_details.rs
```diff
@@ -0,0 +1,64 @@
...
+                account_id, block_height
+            )
+        })?;
+
+    match account_response.kind {
+        QueryResponseKind::ViewAccount(account) => Ok(account),
+        _ => Err(format!(
+            "ViewAccount result expected for {}, block {}",
+            account_id, block_height
+        )),
```
frol, line 62: Well, this should never happen, yet I want you to consider how useful would be the error when you receive it:

> Error: ViewAccount result expected for frol.near, block 100500

My immediate question would be: "AND WHAT? Go ahead and take it for me, why do you bother me?"

src/aggregated/account_details.rs
```diff
@@ -0,0 +1,64 @@
...
+        .map_err(|err| {
+            format!(
+                "Error while delivering ViewAccount for account {}, block {}: {}",
+                account_id, block_height, err
+            )
+        })?
+        .map_err(|_| {
+            format!(
+                "Invalid ViewAccount query for account {}, block {}",
+                account_id, block_height
```
frol, line 53: If the error does not implement Display, it is still better to include the debug version of it:

```suggestion
        .map_err(|err| {
            format!(
                "Invalid ViewAccount query for account {}, block {}: {:?}",
                account_id, block_height, err
```

P.S. I would expect that it implements Display, so we probably can use `{}` instead of `{:?}`

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,182 @@
...
+        .ok_or("`block_height` expected to be u64")?;
+    let total_supply = u128::from_str_radix(&*block.total_supply.to_string(), 10)
+        .map_err(|_| "`total_supply` expected to be u128")?;
+
+    let lockup_ids = accounts::get_lockup_ids_at_block_height(&pool, block_height).await?;
+
+    let mut lockups_locked_tokens: u128 = 0;
+    let mut unfinished_lockups_count: u128 = 0;
+
+    for lockup_id in lockup_ids.iter() {
```
frol, line 125:
```suggestion
    for lockup_account_id in &lockup_account_ids {
```

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::models::Block;
+
+const DAY: std::time::Duration = std::time::Duration::from_secs(86400);
+
+// Compute circulating supply on a daily basis, starting from 13 Oct 2020
+// (Transfers enabled moment on the Mainnet), and put it to the Indexer DB.
+// Circulating supply is calculated by the formula:
+// total_supply - sum(locked_tokens_on_each_lockup) - sum(locked_foundation_account)
+// The value is always computed for the last block in a day (UTC).
+pub(crate) async fn compute_circulating_supply(
```
frol, line 32: I fail to find a good name for this function since it does so many things at once. Some candidates

* `schedule_compute_circulating_supply` (it is not only scheduling, but keeps running)
* `compute_circulating_supply_daily` (it also back-fills, so it is not accurate)
* `aggregate_circulating_supply` :unamused: (it still fills like it is an action, but it is not going to be a single action)
* `handle_circulating_supply_in_background` :unamused:

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,182 @@
...
+) {
+    let retry_duration = DAY.div(12);
+
+    // Adding 10 minutes to be sure that the picture is finalized and we can collect all needed data
+    let ten_minutes: u64 = 10 * 60 * 1000_000_000;
+    let mut day_to_compute =
+        TRANSFERS_ENABLED - TRANSFERS_ENABLED % (DAY.as_nanos() as u64) + ten_minutes;
+
+    loop {
+        time::sleep_until(get_instant_for_timestamp(day_to_compute)).await;
```
frol, line 44: I find this super odd and it took me a few brain-cycles to reason about. `get_instant_for_timestamp` looks suspicious at the first glance.

This whole function tries to be too clever by dealing with back-filling and daily updates in a single place. I don't have immediate suggestions, but I wanted to share the feedback from a person who reads it the first time.

telezhnaya: Thank you
While I was explaining why it's the best solution that we could have here, I created a simpler one. 😄

src/aggregated/account_details.rs
telezhnaya: It doesn't :-( That's why I haven't printed it as I've done 6 lines earlier.
Thank you, added printing with `{:?}`

migrations/2021-08-04-151515_circulating_supply_table/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE aggregated__circulating_supply
+(
+    computed_at_block_timestamp       numeric(20, 0) NOT NULL,
+    computed_at_block_hash            text           NOT NULL,
+    circulating_tokens_supply         numeric(45, 0) NOT NULL,
+    total_tokens_supply               numeric(45, 0) NOT NULL,
+    total_lockup_contracts_count      numeric(45, 0) NOT NULL,
+    unfinished_lockup_contracts_count numeric(45, 0) NOT NULL,
```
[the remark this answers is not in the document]
telezhnaya: Skipped that
True, I'm sure even i32 is enough

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::aggregated::account_details::get_account_balance;
+use crate::aggregated::circulating_supply::lockup;
+use crate::aggregated::circulating_supply::lockup::TRANSFERS_ENABLED;
+use crate::db_adapters::accounts;
+use crate::db_adapters::aggregated::circulating_supply::{
+    add_circulating_supply, get_precomputed_circulating_supply_for_timestamp,
+};
+use crate::db_adapters::blocks::get_latest_block_before_timestamp;
+use crate::models::aggregated::circulating_supply::CirculatingSupply;
+use crate::models::Block;
```
frol, line 25: We prefer to `use crate::models` and then use `models::Block` and such to communicate that the `Block` type is the database model rather than nearcore `Block` or any other block.

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,180 @@
...
+pub(crate) async fn compute_circulating_supply(
+    view_client: Addr<near_client::ViewClientActor>,
+    pool: Database<PgConnection>,
+) {
+    let retry_duration = DAY.div(12);
+
+    // Adding 10 minutes to be sure that the picture is finalized and we can collect all needed data
+    let ten_minutes: u64 = 10 * 60 * 1_000_000_000;
+    let mut day_to_compute =
+        TRANSFERS_ENABLED - TRANSFERS_ENABLED % (DAY.as_nanos() as u64) + ten_minutes;
```
frol, line 44: Can we make it to be `Duration`?

src/models/accounts.rs
```diff
@@ -3,7 +3,8 @@ use bigdecimal::BigDecimal;
 use crate::schema;
 use schema::accounts;
 
-#[derive(Insertable, Debug, Clone)]
+#[table_name = "accounts"]
+#[derive(Insertable, Debug, Clone, QueryableByName)]
```
frol, line 7:
```suggestion
#[derive(Insertable, Debug, Clone, QueryableByName)]
#[table_name = "accounts"]
```

`table_name` attribute becomes available with `QueryableByName` derive, so it makes sense to flip the order

src/main.rs
```diff
@@ -247,23 +250,42 @@ fn main() {
...
                     args.concurrency,
                     args.allow_missing_relations_in_first_blocks,
                 ));
+
+                // Spawning the computation of aggregated data
+                let view_client = indexer.client_actors().0;
+
+                // TODO Is there another way to find out which blockchain do we listen? It's awful
+                let home_dir_str = home_dir.as_path().to_str().unwrap_or("");
+                if home_dir_str.contains("mainnet") {
```
frol, line 283:
```
if indexer.near_config().genesis.config.chain_id == "mainnet" {
```

src/aggregated/circulating_supply/circulating_supply_provider.rs
khorolets: how about `run_circulating_supply_computation`?

khorolets (approved):

src/aggregated/account_details.rs
```diff
@@ -0,0 +1,63 @@
...
+    get_account_view_for_block_height(view_client, account_id, block_height)
+        .await
+        .map(|account| account.code_hash)
+}
+
+async fn get_account_view_for_block_height(
+    view_client: &Addr<ViewClientActor>,
+    account_id: &near_primitives::types::AccountId,
+    block_height: &near_primitives::types::BlockHeight,
+) -> Result<near_primitives::views::AccountView, String> {
```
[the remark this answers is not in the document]
frol: @khorolets very good point, though we struggled with them on the library boundary, and here we are on the application level, so it is less painful, though I would still prefer some error type instead of `String`.

@telezhnaya Given that you are going on vacation soon, and I would prefer to merge this PR before that, I would not dive into error handling just yet. Still, consider reading this great article about error handling in Rust: https://nick.groenen.me/posts/rust-error-handling/. The author has a very good point of various use-cases, and in our case here we should be fine with `anyhow` crate.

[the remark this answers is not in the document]
telezhnaya: https://github.com/near/near-indexer-for-explorer/issues/143

src/aggregated/account_details.rs
telezhnaya: Agree, that was strange. It's still not perfect, but a least with some explanations. Added `map_err` above by stacktrace to explain what exactly went wrong.
Hopefully will do that better on the next iteration in #143

telezhnaya:
@frol @khorolets please have a look

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,193 @@
...
+        .to_u64()
+        .ok_or("`block_timestamp` expected to be u64")?;
+    let block_height = block
+        .block_height
+        .to_u64()
+        .ok_or("`block_height` expected to be u64")?;
+
+    if latest_block_height <= block_height {
+        return Err("Requested block is not finalized".to_string());
+    }
```
frol, line 95: This does not make a lot of sense since you query Postgres for both `latest_block_height` and `block_height` while the local node is what you want to check for the latest available block.

Query view-client for the final block to see what it is on the local node that indexer is running behind the scenes: https://github.com/FroVolod/near-cli/blob/8c37ede9136e13737d9ff0312f04f472877dd4dd/src/commands/view_command/view_account/block_id/mod.rs#L100-L108

P.S. I would probably move this check up to `run_circulating_supply_computation` since this has nothing to do with the circulating supply computation.

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,193 @@
...
+) -> Result<CirculatingSupply, String> {
+    let block_timestamp = block
+        .block_timestamp
+        .to_u64()
+        .ok_or("`block_timestamp` expected to be u64")?;
+    let block_height = block
+        .block_height
+        .to_u64()
+        .ok_or("`block_height` expected to be u64")?;
+    let total_supply = u128::from_str_radix(&*block.total_supply.to_string(), 10)
```
frol, line 139:
```suggestion
    let total_supply = u128::from_str_radix(&block.total_supply.to_string(), 10)
```

src/aggregated/circulating_supply/circulating_supply_provider.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    let circulating_supply: u128 = total_supply - foundation_locked_tokens - lockups_locked_tokens;
+
+    Ok(CirculatingSupply {
+        computed_at_block_timestamp: BigDecimal::from(block_timestamp),
+        computed_at_block_hash: (&block.block_hash).to_string(),
+        circulating_tokens_supply: BigDecimal::from_str(circulating_supply.to_string().as_str())
+            .map_err(|_| "`circulating_tokens_supply` expected to be u128")?,
+        total_tokens_supply: BigDecimal::from_str(total_supply.to_string().as_str())
```
frol, line 182: This is slightly easier to read:

```suggestion
        total_tokens_supply: BigDecimal::from_str(&total_supply.to_string())
```

Alternatively, you may also go with:

```
        total_tokens_supply: total_supply.to_string().parse()
```

src/aggregated/circulating_supply/lockup.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// The lockup contract implementation had a bug that affected lockup start date.
+// For each contract, we should choose the logic based on the binary version of the contract
+pub(crate) fn is_bug_inside_contract(
+    code_hash: &near_primitives::hash::CryptoHash,
+    account_id: &near_primitives::types::AccountId,
+) -> Result<bool, String> {
+    match &*code_hash.to_string() {
+        // The first realization, with the bug
```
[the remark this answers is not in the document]
khorolets: @frol see! @telezhnaya uses the term `realization` too. So sad it's the same word `implementation` in English :)

telezhnaya:
@frol please have a look again

PR #141, opened 2021-08-04
fix: remove CONCURRENTLY from migration

telezhnaya:
CREATE INDEX CONCURRENTLY cannot run inside a transaction block


frol (approved):
I asked how to address it on Diesel discussions board: https://github.com/diesel-rs/diesel/discussions/2841

Let's follow your plan for now and see if we can get any solution from Diesel folks.

PR #142, opened 2021-08-09
feat: introduced an easy way to sort account_changes in their correct historical order (changed_in_block_timestamp + index_in_block)

### LINKED ISSUE #77: Do we miss some balance changes?

telezhnaya:
*Partially resolves #77* (this PR introduces an easy way to sort the account changing events: `ORDER BY changed_in_block_timestamp, index_in_block`)

I am still waiting for the server updates, want to test it on the production-like env, that's why I marked it as a draft.

We should apply the migration by pieces, otherwise, it will take too long to update existing lines.

### DIFF
--- a/migrations/2021-08-06-123500_account_changes_ordering_column/down.sql
+++ b/migrations/2021-08-06-123500_account_changes_ordering_column/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE account_changes DROP COLUMN index_in_block;
--- a/migrations/2021-08-06-123500_account_changes_ordering_column/up.sql
+++ b/migrations/2021-08-06-123500_account_changes_ordering_column/up.sql
@@ -1,0 +1,62 @@
+-- Setting default value -1 for further fill with data to avoid making the field nullable
+ALTER TABLE account_changes
+    ADD COLUMN index_in_block integer NOT NULL DEFAULT -1;
+
+-- This migration is heavy to apply. Consider using Python script to do that in background.
+-- It applies the changes by tiny pieces.
+-- You need to replace `END` and `connection_string` parameters
+
+-- from sqlalchemy import create_engine
+--
+--     START = 1595370903490523743
+--     STEP = 1000 * 1000 * 1000 * 1000  # 1000 secs -> ~17 minutes
+-- END = 1628683241000000000
+-- ESTIMATED_STEPS = (END - START) / STEP
+-- connection_string = 'postgresql+psycopg2://user:pass@host/database'
+--
+--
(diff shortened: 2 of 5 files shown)


frol (approved):

frol:
@telezhnaya how long does it take to apply the migration?

telezhnaya:
@frol the honest answer is "too long anyway". 3 minutes if everything was already updated before.

I think the best idea is to comment `UPDATE` statement in `up.sql` so that we will never apply this migration fully. The comment will be used just to know the idea of how it should be done.
In the real world, we should apply it in pieces. I suggest creating a script that updates the data splitting it by timestamp.

frol:
> I think the best idea is to comment UPDATE statement in up.sql so that we will never apply this migration fully. The comment will be used just to know the idea of how it should be done.
> In the real world, we should apply it in pieces. I suggest creating a script that updates the data splitting it by timestamp.

This sounds like a plan :+1:

khorolets:
@telezhnaya last time we had an issue with "too timeconsuming" migration I've come up with a script and left it for the history here https://github.com/near/near-indexer-for-explorer/pull/99#issuecomment-838871700 you might find it handy. Let me know if you need my help.

frol (approved):

frol:
I just remembered that we may also need an index over the `(changed_in_block_timestamp, index_in_block)` to speed up sorting.

PR #144, opened 2021-08-11
feat: add account changes sorting index

telezhnaya:
#142


frol (approved):
LGMT, but the migration needs to be applied manually with `CONCURRENT` index creation

PR #153, opened 2021-08-23
feat: add ft, nft aggregated tables

telezhnaya:
Please have a look at the columns in both tables.
[FT](https://github.com/near/near-indexer-for-explorer/pull/153/files#diff-dbc27ed818d6f8bbaca70b3e91417c908a5815bfd961e460b75f672955ec70d1R1-R13)
[NFT](https://github.com/near/near-indexer-for-explorer/pull/153/files#diff-dbc27ed818d6f8bbaca70b3e91417c908a5815bfd961e460b75f672955ec70d1R32-R45)

### FT

https://github.com/near/NEPs/issues/141

We can also add nullable `memo` field. Though, I thought it's not so important and we can store it in `args`. I also have a plan to put there all the other data that we have.

### NFT

https://github.com/near/NEPs/discussions/171

Same things about `memo` and `args`.

We also can add nullable `approval_id` field, read more about it here:
https://github.com/near/NEPs/blob/master/specs/Standards/NonFungibleToken/ApprovalManagement.md

Any feedback is highly appreciated!

I guess I can implement the first draft and check what we put into `args`. After that, we can finalize the columns.

### DIFF
--- a/migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
+++ b/migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
@@ -1,62 +1,64 @@
-CREATE TABLE aggregated__fungible_tokens
+CREATE TABLE aggregated__fungible_token_operations
 (
-    id                                bigserial                NOT NULL,
-    included_in_transaction_hash      text                     NOT NULL,
-    included_in_transaction_timestamp numeric(20, 0)           NOT NULL,
-    transaction_status                execution_outcome_status NOT NULL,
-    issued_contract_id                text                     NOT NULL,
-    called_method                     text                     NOT NULL,
-    predecessor_account_id            text                     NOT NULL,
-    receiver_account_id               text                     NOT NULL,
-    amount                            numeric(45, 0)           NOT NULL,
-    args                              jsonb                    NOT NULL
+    processed_in_block_timestamp  numeric(20, 0) NOT NULL,
+    processed_in_transaction_hash text           NOT NULL,
+    processing_index_in_chunk     integer        NOT NULL,
+    ft_contract_account_id        text           NOT NULL,
+    ft_affected_account_id        text           NOT NULL,
+    called_method                 text           NOT NULL,
+    ft_affected_account_balance   numeric(45, 0) NOT NULL,
+    args                          jsonb          NOT NULL
 );
 
-ALTER TABLE ONLY aggregated__fungible_tokens
-    ADD CONSTRAINT aggregated__fungible_tokens_pkey PRIMARY KEY (id);
+ALTER TABLE ONLY aggregated__fungible_token_operations
+    ADD CONSTRAINT aggregated__fungible_token_operations_pkey PRIMARY KEY (processed_in_transaction_hash,
+                                                                           ft_contract_account_id,
+                                                                           ft_affected_account_id);
 
-CREATE INDEX aggregated__fungible_tokens_timestamp_idx ON aggregated__fungible_tokens USING btree (included_in_transaction_timestamp);
+-- I will edit it after we finalise the naming
+-- CREATE INDEX aggregated__fungible_tokens_timestamp_idx ON aggregated__fungible_tokens USING btree (included_in_transaction_timestamp);
+--
+-- ALTER TABLE ONLY aggregated__fungible_tokens
+--     ADD CONSTRAINT aggregated__fungible_tokens_fk FOREIGN KEY (included_in_transaction_hash) REFERENCES transactions (transaction_hash) ON DELETE CASCADE;
+--
+-- ALTER TABLE ONLY aggregated__fungible_tokens
+--     ADD CONSTRAINT aggregated__ft_issued_contract_id_fk FOREIGN KEY (issued_contract_id) REFERENCES accounts (account_id) ON DELETE CASCADE;
+--
+-- ALTER TABLE ONLY aggregated__fungible_tokens
+--     ADD CONSTRAINT aggregated__ft_predecessor_account_id_fk FOREIGN KEY (predecessor_account_id) REFERENCES accounts (account_id) ON DELETE CASCADE;
+--
+-- ALTER TABLE ONLY aggregated__fungible_tokens
+--     ADD CONSTRAINT aggregated__ft_receiver_account_id_fk FOREIGN KEY (receiver_account_id) REFERENCES accounts (account_id) ON DELETE CASCADE;
 
-ALTER TABLE ONLY aggregated__fungible_tokens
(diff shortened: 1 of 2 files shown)


migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
+CREATE TABLE aggregated__fungible_tokens
+(
+    id                                bigserial                NOT NULL,
```
frol, line 3: I feel that `id` is useless. We should have some way to define an order of events, e.g. `processing_index_in_chunk` (`ORDER BY processed_in_block_timestamp, processing_index_in_chunk`  we don't care if tokens have the same `processing_index_in_chunk` since the receipts in the same block do not overlap and even more so, the receipts to the same FT/NFT token contract are bounded to a single shard where the contract resides)

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
...
+    included_in_transaction_hash      text                     NOT NULL,
```
frol, line 4: I suggest `originated_from_transaction_hash` and also consider adding `originated_from_receipt_id`

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
+CREATE TABLE aggregated__fungible_tokens
```
frol, line 1: I feel that "fungible_tokens" is not a self-describing naming. What exactly about FT do we store here? We don't store them as is (like `accounts` or `access_keys` tables), we seem to store updates/touches/changes (like `account_changes` table)

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
...
+(
+    id                                bigserial                NOT NULL,
+    included_in_transaction_hash      text                     NOT NULL,
+    included_in_transaction_timestamp numeric(20, 0)           NOT NULL,
+    transaction_status                execution_outcome_status NOT NULL,
```
frol, line 6: Will we actually set the **transaction** status? Can we actually say if the FT operation was successful as a whole transaction rather than just a single receipt execution outcome status? (I have yet to read the FT standard, so keep that in mind) How would that play with cross-contract calls (e.g. FT operation is triggered as part of some other transaction just as a cross-contract call)

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
...
+    issued_contract_id                text                     NOT NULL,
```
frol, line 7: It might make sense, but from a first glance I cannot say what it is, so it might be worth considering a better name (I will need to read FT NEP before I can provide any suggestions)

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
```diff
@@ -0,0 +1,62 @@
...
+    called_method                     text                     NOT NULL,
+    predecessor_account_id            text                     NOT NULL,
+    receiver_account_id               text                     NOT NULL,
```
frol, line 10: From the "fungible token" table point of view these are ambiguous names since `receiver_account_id` of a transaction, a receipt, and FT token are three different account ids, and I feel that in the context of "fungible token" table we care about the FT sender and FT receiver account ids and the rest is less relevant.

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
telezhnaya: I am OK with adding `processing_index_in_chunk` for sorting purposes, but I'm still not sure is it a good idea to get rid of the artificial `id` column. 

Some thoughts on why having `id` as PK is good:
- I still think about duplicated tx hashes and maybe it's better to gradually migrate to artificial `id` as PK everywhere. In my opinion, our first task is to show data, only after it we can validate it somehow. So, if `timestamp` + `index_in_chunk` is not unique for some reason, it's better to add the value anyway, and then decide what should we do with it.
- If the PK is large, you have fewer keys per page in the index and the index will take more cache memory to store it. We have 8 bytes (one id bigserial) VS 20 bytes (16 bytes in timestamp numeric(20,0) and 4 bytes in index_in_chunk integer). Performance is also degraded because we have to scan more memory.
- Comparing numerics could be more expensive than comparing integers. I'm not sure about it, it is not mentioned in the documentation in a straight way, but 5-minute googling gives this thought.

migrations/2021-08-23-161500_aggregated_ft_nft_tables/up.sql
telezhnaya:
```
aggregated__fungible_token_actions
aggregated__fungible_token_updates
aggregated__fungible_token_changes
aggregated__fungible_token_operations
```

I vote for `operations` and `changes`

frol:
I feel that receipt-based records will be much easier to reason about and won't break apart if there is a contract that once being called initiates two independent FT transfers (being a single transaction, it will be hard to store this kind of scenario in transaction-based records); also, waiting for all the receipts sounds too much of a hassle on the indexer level since we process data block-by-block.

From the usage perspective, I would love to be able to:

```sql
SELECT
  ft_sender_account_id,
  ft_receiver_account_id,
  ft_name, -- ambiguous name, so we should do better
  ft_amount, -- UNSURE if we should have a transfer amount or should we store the updated balance or both (I feel that the transfer amount is more useful and the balance can be fetched on-demand via view call given we know the block reference; though we have a minor issue if a few transfers land on the same block, then we need to fetch the balance on a previous block, and all the transfers on the current block in the correct order and apply the changes correctly)
  originated_from_transaction_hash,
  updated_at_receipt_id,
FROM aggregated__fugnible_tokens_operations
WHERE ft_sender_account_id = 'frol.near' OR ft_receiver_account_id = 'frol.near'
ORDER BY updated_at_block_timestamp DESC, updated_at_index_in_chunk DESC
LIMIT 10
```

In order to see:

```
frol.near history:

frol.near sent 10 wNEAR to olga.near | frol.near's balance is 90 wNEAR:
  [deadbeef01] frol.near has been deducted 10 wNEAR to send to olga.near
  [deadbeef02] olga.near has received 10 wNEAR from frol.near
frol.near received 7 wNEAR from khorolets.near | frol.near's balance is 97 wNEAR
  ...
```

telezhnaya:
We merged NFT support recently #169; it's easier for me to start from the beginning with FT support rather than using this discussion. Closing the PR

PR #154, opened 2021-08-23
fix: Make diesel schema generator work again

khorolets:
`src/schema.rs` is generated by command `diesel print-schema src/schema.rs`

It was a little bit broken because of the missing table in `diesel.toml`. Also, @telezhnaya has introduced `aggregated__lockups` view in SQL and we need it as a `table!` in schema. To avoid doing it manually I've made use of `patch_file` field from [`diesel-orm` documentation](http://diesel.rs/guides/configuring-diesel-cli.html)

In addition, it removes unused imports which I would do manually for a year.


src/schema.patch
```diff
@@ -0,0 +1,131 @@
+diff --git a/src/schema.rs b/src/schema.rs
+index d0d6a3b..9d4f5d1 100644
+--- a/src/schema.rs
++++ b/src/schema.rs
+@@ -30,13 +30,12 @@ table! {
```
telezhnaya, line 5: Let's add the comment here, something like "this file is autogenerated, consider edit diesel.toml"?

Sorry again that I edited `schema.rs` manually 🙂

telezhnaya (approved):
Thank you!

I think it's a good idea to edit instructions in README (running or local debugging) by adding step with
 `diesel print-schema src/schema.rs`

PR #155, opened 2021-08-23
feat: Upgrade to 0.9.3 (NEAR Indexer Framework 0.10.0, refactor storing genesis records, more logs on indexer start)

khorolets:
It should help to ensure the indexer is working and is going to start streaming and has not stuck

Also, bumped the version to 0.9.3 because of:
* upgrade to NEAR Indexer Framework 0.10.0
* change the logic of storing genesis records
* add more logs on indexer start
* include `escape_json` function to serializers

### DIFF
--- a/src/db_adapters/genesis.rs
+++ b/src/db_adapters/genesis.rs
@@ -24,19 +24,23 @@ pub(crate) async fn store_genesis_records(
 
     near_config.genesis.for_each_record(|record| {
         if accounts_to_store.len() == 5_000 {
-            let accounts_to_store_copy = accounts_to_store.clone();
+            let mut accounts_to_store_chunk = vec![];
+            std::mem::swap(&mut accounts_to_store, &mut accounts_to_store_chunk);
             let pool_copy = pool.clone();
-            accounts_to_store.clear();
+            // This will block current thread but there is no other async tasks
+            // running at the moment so it is fine to cut this corner for now
             tokio_runtime.block_on(async move {
-                store_accounts_from_genesis(pool_copy, accounts_to_store_copy).await;
+                store_accounts_from_genesis(pool_copy, accounts_to_store_chunk).await;
             });
         }
         if access_keys_to_store.len() == 5_000 {
-            let access_keys_to_store_copy = access_keys_to_store.clone();
+            let mut access_keys_to_store_chunk = vec![];
+            std::mem::swap(&mut access_keys_to_store, &mut access_keys_to_store_chunk);
             let pool_copy = pool.clone();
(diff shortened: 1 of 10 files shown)


frol:
I tested it with testnet genesis (5.4GB JSON file) on my local machine with local postgres database and it works great!

* peak RAM usage was 4.6GB, but it does not seem to be related to the indexer since while the genesis processing was happening, the RAM usage dropped down to 3.8GB and never go up
* time-wise it looks great!
  * 10 minutes to "Load config from" (not Indexer problem, nearcore does something heavy there)
  * 35 seconds to store accounts from genesis
  * 60 seconds to skip unrelated genesis records
  * 60 seconds to store access keys from genesis
* nearcore is not blocked by this storing the genesis data, which means that the implementation is async-friendly

src/main.rs
```diff
@@ -219,7 +220,7 @@ fn main() {
     let pool = models::establish_connection();
 
     let mut env_filter = EnvFilter::new(
-        "tokio_reactor=info,near=info,near=error,stats=info,telemetry=info,indexer_for_explorer=info,aggregated=info",
+        "tokio_reactor=info,near=info,near=error,stats=info,telemetry=info,indexer=info,indexer_for_explorer=info,aggregated=info",
```
telezhnaya, line 223: I can't find usages of this, do we need it?

What is the difference between "indexer" and "indexer_for_explorer"?

src/db_adapters/genesis.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        near_config.genesis.for_each_record(|record| {
+            if accounts_to_store.len() == 5_000 {
+                let mut accounts_to_store_chunk = vec![];
+                std::mem::swap(&mut accounts_to_store, &mut accounts_to_store_chunk);
+                let pool = pool.clone();
+                block_on(actix_arbiter, async move {
+                    store_accounts_from_genesis(pool, accounts_to_store_chunk).await;
+                })
+                .expect("storing accounts from genesis failed");
```
telezhnaya, line 71: The error is incomplete, storing access keys probably also failed.

telezhnaya (approved):
I didn't fully catch the idea of `block_on` function, I will ask about it today on the call 🙂

src/db_adapters/genesis.rs
khorolets: didn't get, could you explain?

src/main.rs
khorolets: `indexer` is NEAR Indexer (from nearcore) and `indexer_for_explorer` is current one

PR #156, opened 2021-08-24
fix: cleanup in circulating supply

telezhnaya:
move run_circulating_supply_computation to mod rs
prepare the place for other aggregated modules
get rid of cargo clippy warnings

PR #157, opened 2021-08-24
fix: use anyhow errors in circulating supply

telezhnaya:
Look at this after #156


src/aggregated/account_details.rs
```diff
@@ -7,29 +8,32 @@ pub(crate) async fn get_account_balance(
     view_client: &Addr<ViewClientActor>,
     account_id: &near_primitives::types::AccountId,
     block_height: &near_primitives::types::BlockHeight,
-) -> Result<near_primitives::types::Balance, String> {
+) -> anyhow::Result<near_primitives::types::Balance> {
     get_account_view_for_block_height(view_client, account_id, block_height)
         .await
         .map(|account| account.amount)
-        .map_err(|err| format!("Unable to get account balance: {}", err))
+        .context(format!("Unable to get account balance for {}", account_id))
```
frol, line 15: It is better use `.with_context()`, so we don't trigger `format!` when error does not occur.

https://docs.rs/anyhow/1.0.43/anyhow/trait.Context.html

frol (approved):

PR #162, opened 2021-09-06
fix: Add `escape_json` function to the serializer

khorolets:


src/models/serializers.rs
```diff
@@ -138,3 +139,20 @@ pub(crate) fn extract_action_type_and_value_from_action_view(
         ),
     }
 }
+
+fn escape_json(object: &mut serde_json::Value) {
```
frol, line 143: Let's add a comment here why this function is necessary (NULL-bytes limitation of PostgreSQL in TEXT  and JSONB columns) and reference this post: https://www.commandprompt.com/blog/null-characters-workarounds-arent-good-enough/

frol (approved):

PR #164, opened 2021-09-07
refactor: Strict/Non-strict mode improvements

### LINKED ISSUE #163: Fix `--allow-missing-relations-in-first-blocks`
Currently, `account_changes` constraints cast this feature to a useless one.

Plan to adjust the non-strict mode to:

- [ ] turn off `account_changes` handler in the non-strict mode for the number of blocks provided
- [ ] decrease the number of retries to be performed in non-strict mode (currently every retry increases the delay between tries)

This should allow us to start indexer from any block to speed up catching up or something similar in emergency situations.

khorolets:
Fixes #163 

I've bumped the version to `0.10.0` because I've dropped the `--allow-missing-relations-in-first-blocks` flag completely. It **doesn't mean** this PR is going to be released once merged.

* Dropped `--allow-missing-relations-in-first-blocks` flag completely
* Add the flag `--non-strict-mode` which: 
  * changes retry attempts on Receipts from 10 to 4
  * stops saving `account_changes` completely as it is impossible to save in case of missing of related Receipts
* Add the flag `--stop-after-number-of-blocks N` which namely stops the indexer after processing N blocks

All those flags and features are useful for debugging or emergency purposes. They are not supposed to be used in normal circumstances.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Drop `--allow-missing-relations-in-first-blocks` flag
 * Introduce `--non-strict-mode` which does the same as `--allow-missing-relations-in-first-blocks` flag did but endlessly
-* Add `--stop-after-number-of-blocks <amount>` flag to plan Indexer for Explorer to stop once it indexed the provided `<amount>` of blocks. May be useful for debug or maintenance purposes.
+* Add `--stop-after-number-of-blocks <count>` flag to plan Indexer for Explorer to stop once it indexed the provided `<count>` of blocks. May be useful for debug or maintenance purposes.
 
 ## Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.10.0
+
+* Drop `--allow-missing-relations-in-first-blocks` flag
+* Introduce `--non-strict-mode` which does the same as `--allow-missing-relations-in-first-blocks` flag did but endlessly
+* Add `--stop-after-number-of-blocks <amount>` flag to plan Indexer for Explorer to stop once it indexed the provided `<amount>` of blocks. May be useful for debug or maintenance purposes.
+
+## Breaking changes
+
+* The flag `--allow-missing-relations-in-first-blocks` is not available anymore in favor of `--non-strict-mode` flag
+
 ## 0.9.3
 
 * Escape `args_json` on the fly to avoid null-byte issues
--- a/src/configs.rs
+++ b/src/configs.rs
@@ -38,12 +38,12 @@ pub(crate) struct RunArgs {
     /// Force streaming while node is syncing
     #[clap(long)]
     pub stream_while_syncing: bool,
-    /// Switches indexer to non-strict mode (skips Receipts without parent Transaction hash, stops storing AccountChanges)
+    /// Switches indexer to non-strict mode (skips Receipts without parent Transaction hash, stops storing AccountChanges and AccessKeys)
     #[clap(long)]
     pub non_strict_mode: bool,
     /// Stops indexer completely after indexing the provided number of blocks
     #[clap(long, short)]
-    pub stop_after_number_of_blocks: Option<u64>,
+    pub stop_after_number_of_blocks: Option<std::num::NonZeroUsize>,
     /// Sets the concurrency for indexing. Note: concurrency (set to 2+) may lead to warnings due to tight constraints between transactions and receipts (those will get resolved eventually, but unless it is the second pass of indexing, concurrency won't help at the moment).
     #[clap(long, default_value = "1")]
     pub concurrency: std::num::NonZeroU16,
--- a/src/configs.rs
+++ b/src/configs.rs
@@ -38,12 +38,17 @@ pub(crate) struct RunArgs {
     /// Force streaming while node is syncing
     #[clap(long)]
     pub stream_while_syncing: bool,
+    /// Switches indexer to non-strict mode (skips Receipts without parent Transaction hash, stops storing AccountChanges)
     #[clap(long)]
-    pub allow_missing_relations_in_first_blocks: Option<u32>,
-    #[clap(subcommand)]
-    pub sync_mode: SyncModeSubCommand,
+    pub non_strict_mode: bool,
+    /// Stops indexer completely after indexing the provided amount of blocks
+    #[clap(long, short)]
+    pub stop_after_number_of_blocks: Option<u64>,
+    /// Sets the concurrency for indexing. Note: high concurrency level may lead to race conditions
     #[clap(long, default_value = "1")]
     pub concurrency: std::num::NonZeroU16,
+    #[clap(subcommand)]
+    pub sync_mode: SyncModeSubCommand,
 }
 
 #[allow(clippy::enum_variant_names)] // we want commands to be more explicit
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.10.0
+
+* Drop `--allow-missing-relations-in-first-blocks` flag
+* Introduce `--non-strict-mode` which does the same as `--allow-missing-relations-in-first-blocks` flag did but endlessly
+* Add `--stop-after-number-of-blocks <amount>` flag to plan Indexer for Explorer to stop once it indexed the provided `<amount>` of blocks. May be useful for debug or maintenance purposes.
```
frol, line 7:
```suggestion
* Add `--stop-after-number-of-blocks <count>` flag to plan Indexer for Explorer to stop once it indexed the provided `<count>` of blocks. May be useful for debug or maintenance purposes.
```

Use `amount` for non-individually countable nouns (mass, money, tokens, etc) and `count`/`number` for individually countable (blocks, transactions, people)

src/configs.rs
```diff
@@ -38,12 +38,17 @@ pub(crate) struct RunArgs {
     /// Force streaming while node is syncing
     #[clap(long)]
     pub stream_while_syncing: bool,
+    /// Switches indexer to non-strict mode (skips Receipts without parent Transaction hash, stops storing AccountChanges)
     #[clap(long)]
-    pub allow_missing_relations_in_first_blocks: Option<u32>,
-    #[clap(subcommand)]
-    pub sync_mode: SyncModeSubCommand,
+    pub non_strict_mode: bool,
+    /// Stops indexer completely after indexing the provided amount of blocks
```
frol, line 44: See my other comment about "amount"

src/configs.rs
```diff
@@ -38,12 +38,17 @@ pub(crate) struct RunArgs {
...
+    #[clap(long, short)]
+    pub stop_after_number_of_blocks: Option<u64>,
+    /// Sets the concurrency for indexing. Note: high concurrency level may lead to race conditions
```
frol, line 47:
```suggestion
    /// Sets the concurrency for indexing. Note: concurrency (set to 2+) may lead to warnings due to tight constraints between transactions and receipts (those will get resolved eventually, but unless it is the second pass of indexing, concurrency won't help at the moment).
```

src/configs.rs
```diff
@@ -59,7 +64,7 @@ pub(crate) enum SyncModeSubCommand {
 
 #[derive(Clap, Debug, Clone)]
 pub(crate) struct InterruptionArgs {
-    /// start sycing this amount of blocks earlier than actual interruption
+    /// start syncing this amount of blocks earlier than actual interruption
```
frol, line 67:
```suggestion
    /// start indexing this number of blocks earlier than the actual interruption happened
```

src/main.rs
```diff
@@ -96,38 +96,56 @@ async fn handle_message(
...
         .enumerate()
         .map(|(index, streamer_message)| {
+            if let Some(index_to_stop) = stop_after_number_of_blocks {
+                if (index_to_stop) as usize == index {
+                    info!(
+                        target: crate::INDEXER_FOR_EXPLORER,
+                        "Stopping after according to `--stop-after-number-of-blocks {}`",
+                        index_to_stop,
+                    );
+                    std::process::exit(0);
```
frol, line 141: CRITICAL: This might corrupt the database. We should shut nearcore down gracefully (ask @miraclx if you need help)

src/main.rs
```diff
@@ -96,38 +96,56 @@ async fn handle_message(
...
+    strict_mode: bool,
+    stop_after_number_of_blocks: Option<u64>,
 ) {
     tracing::info!(target: crate::INDEXER_FOR_EXPLORER, "Stream has started");
-    let strict_mode = allow_missing_relation_in_start_blocks.unwrap_or(0);
     let mut handle_messages = tokio_stream::wrappers::ReceiverStream::new(stream)
         .enumerate()
         .map(|(index, streamer_message)| {
+            if let Some(index_to_stop) = stop_after_number_of_blocks {
+                if (index_to_stop) as usize == index {
```
frol, line 135: Can we just use `.take(stop_after_number_of_blocks)` on the iterator? This will gracefully handle all the messages in the `while` loop below and just exit from Indexing (we will only need to gracefully shut down nearcore after we complete here).

frol (changes requested):

src/configs.rs
khorolets:
```suggestion
    /// Stops indexer completely after indexing the provided number of blocks
```

src/main.rs
```diff
@@ -1,6 +1,5 @@
-use std::convert::{TryFrom, TryInto};
-
 use clap::Clap;
+use std::convert::{TryFrom, TryInto};
```
frol, line 2: Why? std imports are more generic, so should be above

src/main.rs
```diff
@@ -96,42 +95,59 @@ async fn handle_message(
...
                 "Block height {}", &streamer_message.block.header.height
             );
-            handle_message(&pool, streamer_message, index >= strict_mode as usize)
-        })
-        .buffer_unordered(usize::from(concurrency.get()));
+            handle_message(&pool, streamer_message, strict_mode)
+        });
+    let mut handle_messages = if let Some(stop_after_n_blocks) = stop_after_number_of_blocks {
+        handle_messages
+            .take(stop_after_n_blocks as usize)
```
frol, line 140: I would either make `stop_after_number_of_blocks` to be `usize` in the first place, or use `usize::try_from().expect()` since `as usize` might cause unexpected issues on non-64-bit platforms down the road.

src/configs.rs
```diff
@@ -38,12 +38,17 @@ pub(crate) struct RunArgs {
...
     pub stream_while_syncing: bool,
+    /// Switches indexer to non-strict mode (skips Receipts without parent Transaction hash, stops storing AccountChanges)
     #[clap(long)]
-    pub allow_missing_relations_in_first_blocks: Option<u32>,
-    #[clap(subcommand)]
-    pub sync_mode: SyncModeSubCommand,
+    pub non_strict_mode: bool,
+    /// Stops indexer completely after indexing the provided number of blocks
+    #[clap(long, short)]
+    pub stop_after_number_of_blocks: Option<u64>,
```
frol, line 46: Should it be NonZero?

src/main.rs
```diff
@@ -96,42 +95,59 @@ async fn handle_message(
...
+            .boxed_local()
+    } else {
+        handle_messages.boxed_local()
+    }
+    .buffer_unordered(usize::from(concurrency.get()));
 
     while let Some(_handled_message) = handle_messages.next().await {}
+    // Graceful shutdown
+    drop(handle_messages);
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
frol, line 150: Let's have a log here that the graceful shutdown will happen in 5 seconds, so whoever uses it, don't think it got stuck.

src/main.rs
```diff
@@ -96,42 +95,59 @@ async fn handle_message(
...
+            execution_outcomes_future,
+            accounts_future,
+            access_keys_future,
+            account_changes_future,
+        );
+    } else {
+        join!(
+            execution_outcomes_future,
+            accounts_future,
+            access_keys_future,
```
frol, line 117: Storing access keys may also fail in non-strict mode, so we should skip them.

P.S. Update the Clap help to indicate this.

frol (approved):
Overall, looks good to me!

frol (approved):

PR #168, opened 2021-09-24
refactor: Retry or panic macros to replace code duplication and add panic after N attempts

khorolets:
WIP

### DIFF
--- a/src/db_adapters/blocks.rs
+++ b/src/db_adapters/blocks.rs
@@ -2,7 +2,6 @@ use actix_diesel::dsl::AsyncRunQueryDsl;
 use anyhow::Context;
 use bigdecimal::{BigDecimal, ToPrimitive};
 use diesel::{ExpressionMethods, PgConnection, QueryDsl};
-use tracing::error;
 
 use near_indexer::near_primitives;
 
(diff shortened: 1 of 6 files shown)


src/retriable.rs
```diff
@@ -0,0 +1,32 @@
...
+        loop {
+            retry_attempt += 1;
+            if retry_attempt == $number_of_retries {
+                panic!(
+                    "Failed to perform query to database after {} attempts. Stop trying.",
+                    $number_of_retries
+                );
+            }
+            match $query.await {
+                Ok(_) => {}
```
frol, line 15: I feel it should `break result;` here, otherwise it is an endless loop here

src/db_adapters/blocks.rs
```diff
@@ -16,30 +15,15 @@ pub(crate) async fn store_block(
...
-                );
-                tokio::time::sleep(interval).await;
-                if interval < crate::MAX_DELAY_TIME {
-                    interval *= 2;
-                }
-            }
-        }
-    }
+            .execute_async(&pool),
+        100,
```
frol, line 23: Adjust this number so we don't stuck for more than 5 minutes. By that time we should be notified about the failure.

src/retriable.rs
```diff
@@ -0,0 +1,32 @@
+#[macro_export]
+macro_rules! execute_retriable_or_panic {
```
frol, line 2: I feel it is generic enough to drop "execute". Maybe `await_and_retry_or_panic`? Ugly names for macro solutions :smile:  Please, think of a better name :pray:

frol (approved):

PR #169, opened 2021-10-05
feat: add NFT events support

telezhnaya:
I mentioned most of the things in the doc inside the SQL script, please have a look at that. 
Additionally, I want to help the contract creators to find the relevant information.
We discussed the format of event logs [here](https://github.com/near/NEPs/issues/254#issuecomment-927729122)

From the event log, we (as Indexer developers) should be able to retrieve the following information:
* `token_id`
* `event_kind` (mint, transfer, burn), reminder: we decided to group by this
* `token_sender_account_id` (should be skipped for mint)
* `token_receiver_account_id` (should be skipped for burn)
* `token_transfer_approval_account_id` (optional, see the info below, **TODO** we haven't discussed that)
* `token_transfer_memo` (optional, feel free not to fill it inside your batch event log if it's empty or you are worried about log size)

#### `token_transfer_approval_account_id`
https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html
You have to fill in this information in the event log if it's relevant for the current transfer. 
Let's looks at the example:
If Alice approves Bob to transfer her token, and then Bob transfers it to Maggie, then the fields should look like this:
```
token_sender_account_id : alice.near
token_receiver_account_id : maggie.near
token_transfer_approval_account_id : bob.near
```
The same situation, but Alice as the owner does the transfer by herself:
```
token_sender_account_id : alice.near
token_receiver_account_id : maggie.near
token_transfer_approval_account_id : <empty>
```

We DO NOT store `called_method` anywhere because we get the info from logs, the method could be retrieved from receipt if you need. @MaximusHaximus do you need this info in Wallet?

@frol @MaximusHaximus @mikedotexe
@evergreen-trading-systems @zcstarr @robert-zaremba


telezhnaya:
FYI: the discussion could be both here or after https://github.com/near/NEPs/issues/254#issuecomment-934487778

evergreen-trading-systems:
awesome. why do we want to add permissions to log authors? because this is new. even if we wanted to, it's impossible to check permissions based on logs that anyone can send.

telezhnaya:
@evergreen-trading-systems it will be a complete anarchy. Without simple check, anyone can write any contract with any logs and make fake transfers, and we will believe these logs. Even if noone is interested in it because of financial reasons, be sure that we will have a disaster on each April 1st. This is too easy to hack. 
I agree that we can trust the authors of the contract, but we should distinguish somehow between the author and unknown contract.

telezhnaya:
UPD: `emitted_by_contract_account_id` could be always taken from receipt_receiver_account_id, and that would be the natural way of checking the author of logs. Updated the message above

src/main.rs
```diff
@@ -64,6 +64,9 @@ async fn handle_message(
         }
     }
 
+    // Assets (NFT)
+    db_adapters::assets::non_fungible_token_events::store_nft(&pool, &streamer_message).await;
```
frol, line 68: Instead of awaiting immediately, create a future and add it to the `join!`s below, so we can do things in parallel.

src/schema.patch
```diff
@@ -76,8 +86,22 @@ index d0d6a3b..8928a6e 100644
...
+ 
++    #[allow(non_snake_case)]
+     assets__non_fungible_token_events (emitted_for_receipt_id, emitted_at_block_timestamp, emitted_in_shard_id, emitted_index_of_event_entry_in_shard, emitted_by_contract_account_id, token_id, event_kind, token_old_owner_account_id, token_new_owner_account_id, token_authorized_account_id, token_memo) {
+         emitted_for_receipt_id -> Text,
+         emitted_at_block_timestamp -> Numeric,
+         emitted_in_shard_id -> Numeric,
+         emitted_index_of_event_entry_in_shard -> Int4,
+         emitted_by_contract_account_id -> Text,
+@@ -126,13 +137,23 @@ table! {
+         token_memo -> Text,
```
frol, line 104: The table (this whole `assets__non_fungible_token_events` block) should not be part of the patch

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -0,0 +1,196 @@
...
+                                emitted_at_block_timestamp: BigDecimal::from(*block_timestamp),
+                                emitted_in_shard_id: BigDecimal::from(*shard_id),
+                                emitted_index_of_event_entry_in_shard: *index_in_shard,
+                                emitted_by_contract_account_id: contract_id.to_string(),
+                                token_id: token_id.to_string(),
+                                event_kind: models::enums::NftEventKind::Mint,
+                                token_old_owner_account_id: "".to_string(),
+                                token_new_owner_account_id: mint_event.owner_id.to_string(),
+                                token_authorized_account_id: "".to_string(),
+                                token_memo: memo.to_string(),
```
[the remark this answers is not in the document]
frol: I don't see a better way either.

P.S. It seems you don't need most of those `.to_string()`, do you?

frol (commented):
Thanks for thoroughly documenting the data schema!

migrations/2021-10-04-100000_assets_nft/up.sql
```diff
@@ -0,0 +1,82 @@
...
+    -- New owner of the token. Empty if we have nft_event_kind 'BURN'.
+    token_new_owner_account_id            text           NOT NULL DEFAULT '',
+    -- The account that initialized the event.
+    -- It differs from token_old_owner_account_id, but it is approved to manipulate with current token.
+    -- More information here https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html
+    -- Optional field: filled only if the event is done NOT by token_old_owner_account_id.
+    -- Empty if we have nft_event_kind 'MINT'.
+    token_authorized_account_id           text           NOT NULL DEFAULT '',
+    -- Optional message associated with token movement.
+    token_memo                            text           NOT NULL DEFAULT ''
```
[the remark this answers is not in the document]
telezhnaya: Agree, it's more about the event than the token. I'm OK with both options, will change it to `event_memo`

src/db_adapters/assets/non_fungible_token_events.rs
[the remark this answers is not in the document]
telezhnaya: It's possible to emit it only in `token_id`.
In other lines, we have to copy the value, because we use them in cycle and move on each iteration. To say that explicitly, I will change it to `.clone()`

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -0,0 +1,196 @@
...
+                    "Error occurred while NonFungibleTokenEvent were adding to database. Retrying in {} milliseconds... \n {:#?} \n{:#?}",
+                    interval.as_millis(),
+                    async_error,
+                    &nft_events,
+                );
+                    tokio::time::sleep(interval).await;
+                    if interval < crate::MAX_DELAY_TIME {
+                        interval *= 2;
+                    }
+                }
```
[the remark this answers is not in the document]
telezhnaya: We can if we extend it with a parameter with a custom function for error handling.
I always use Rust macro by intuition. I anyway wanted to read about macro. If it's not hard, I will do that 🙂

src/schema.patch
telezhnaya: It's not the part of the patch, only `+    #[allow(non_snake_case)]` is

telezhnaya:
@frol sorry for such a massive number of changed files: `cargo clippy` becomes stricter, as I understand.

Important things since your last check:
- I fixed your comments, it's obvious 🙂 
- I rewrote `src/retriable.rs` so that it's possible to pass custom error handling function
- I fixed a bug in `await_retry_or_panic`, it invoked the logic 1 times less than expected. If you invoked it with 1 attempt, it gave you an immediate error.
- I fixed all `cargo clippy` warnings

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -0,0 +1,187 @@
...
+                    "NFT: provided event log does not correspond to any of formats defined in NEP. Will ignore this event. \n {:#?} \n{:#?}",
+                    err,
+                    untrimmed_log,
+                );
+                continue;
+            }
+        };
+
+        let nft_types::NearEvent::Nep171(nep171_event) = event;
+        event_logs.push(nep171_event);
```
frol, line 104: *It is not that important in this case, but I wanted to share this piece of knowledge for the future.*

When you create a vector only to later iterate over it, you may consider having an iterator like:

```rust
let event_logs = outcome.execution_outcome.outcome.logs.filter_map(|untrimmed_log| {
    if ... {
        return None;  // instead of `continue`
    }
    return Some(nep171_event)
});
```

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -0,0 +1,187 @@
...
+                                emitted_at_block_timestamp: BigDecimal::from(*block_timestamp),
+                                emitted_in_shard_id: BigDecimal::from(*shard_id),
+                                emitted_index_of_event_entry_in_shard: *index_in_shard,
+                                emitted_by_contract_account_id: contract_id.to_string(),
+                                token_id,
+                                event_kind: models::enums::NftEventKind::Mint,
+                                token_old_owner_account_id: "".to_string(),
+                                token_new_owner_account_id: mint_event.owner_id.clone(),
+                                token_authorized_account_id: "".to_string(),
+                                event_memo: memo.clone(),
```
frol, line 127: I just realized that most of these values (memo, token_id, some of the account_ids) are actually user-driven, and we have already bumped into the problems of people dumping garbage there, and when it is NULL-byte there, PostgreSQL is not happy about it at all, so we had to escape the strings before storing to the database:

https://github.com/near/near-indexer-for-explorer/blob/0e4fe0cb40fa0552ec97093a1f8c90313ef5b2ec/src/models/serializers.rs#L88

src/schema.patch
```diff
@@ -76,15 +86,29 @@ index d0d6a3b..8928a6e 100644
...
+         emitted_index_of_event_entry_in_shard -> Int4,
+         emitted_by_contract_account_id -> Text,
+@@ -124,15 +135,24 @@ table! {
+         token_new_owner_account_id -> Text,
+         token_authorized_account_id -> Text,
+         event_memo -> Text,
+     }
+ }
+ 
++table! {
```
frol, line 110: I feel this should not be part of the patch. The table definition should be generated automatically.

frol (approved):

src/schema.patch
telezhnaya: It's not the table, but the view, it's not supported by diesel, as I understand

src/db_adapters/assets/non_fungible_token_events.rs
telezhnaya: Thank you!

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -0,0 +1,202 @@
...
+                    err,
+                    untrimmed_log,
+                );
+                return None;
+            }
+        };
+
+        let nft_types::NearEvent::Nep171(nep171_event) = event;
+        Some(nep171_event)
+    }).collect();
```
frol, line 104: Just FYI, my point was to avoid Vec. You should be able to just drop .collect() (and drop the explicit type) and that will give you an iterator, so you can directly use it in the loop

telezhnaya: Oh, now I see.
I was thinking about `map` function in the next piece of code, but the problem was that each event can produce multiple DB rows, and I don't know how to write it in a functional style in Rust.

I was happy to change at least the first cycle here, but I didn't finish it properly 😄 
I will fix it in another small PR!

PR #177, opened 2021-10-12
Add Dockerfile and .dockerignore

mieubrisse:


khorolets:
Hey @mieubrisse! Thanks for the PR! 

Could you clarify have you considered using [multi-stage](https://docs.docker.com/develop/develop-images/multistage-build/) Docker builds?

I'll leave a few review comments soon.

khorolets:
@mieubrisse yeah. My concern is that the current approach is a bit suboptimal for production use because migration step and `--store-genesis` are needed ideally only once. OK, the migration step might be needed from time to time, but still.

How about refactoring this to a multi-stage build? I believe we can find a ready-one Docker image with `diesel-cli` already and describe how to use it, what do you think?

frol:
> how did you do that? Just install libqp-dev and copy the diesel binary across?

Yep

mieubrisse:
Any idea what the other packages besides `libpq-dev` are that are needed, @frol ? I can spend a little while giving a multistage build a go, but the build takes 45min so getting it wrong really hurts (though I tried to be clever and push the build as early in the Dockefile as possible) 

From https://github.com/near/nearcore/blob/master/Dockerfile I also have `libssl-dev` and `ca-certificates`... any others that I need?

Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+WORKDIR /near/indexer-explorer
+
+COPY --from=build /usr/local/cargo/bin/diesel .
+COPY --from=build /tmp/target/release/indexer-explorer .
+# Diesel needs a migrations directory to run
+COPY --from=build /near/indexer-explorer/migrations ./migrations
+ 
+# If the --store-genesis flag isn't set, the accounts in genesis won't get created in the DB which will lead to foreign key constraint violations
+# See https://github.com/near/near-indexer-for-explorer/issues/167
+CMD ./diesel migration run && \
```
khorolets, line 73: Have you considered exposing only the binary here, so users of the image could pass their own args?

Example

```
docker run -it near/indexer-for-explorer --home-dir <...> init ...
```

PR #203, opened 2021-12-01
Handle collisions in transactions table

### LINKED ISSUE #84: Some rows are missing because of transaction hash collisions

telezhnaya:
This PR will hopefully close #84 in one day

### DIFF
--- a/src/db_adapters/receipts.rs
+++ b/src/db_adapters/receipts.rs
@@ -20,7 +20,6 @@ pub(crate) async fn store_receipts(
     block_timestamp: u64,
     strict_mode: bool,
 ) -> anyhow::Result<()> {
-    let block_hash_str = block_hash.to_string();
     let futures = shards
         .iter()
         .filter_map(|shard| shard.chunk.as_ref())
@@ -29,7 +28,7 @@ pub(crate) async fn store_receipts(
             store_chunk_receipts(
                 pool,
                 &chunk.receipts,
-                &block_hash_str,
+                block_hash,
                 &chunk.header.chunk_hash,
                 block_timestamp,
                 strict_mode,
@@ -42,7 +41,7 @@ pub(crate) async fn store_receipts(
 async fn store_chunk_receipts(
     pool: &actix_diesel::Database<PgConnection>,
     receipts: &[near_indexer::near_primitives::views::ReceiptView],
-    block_hash: &str,
+    block_hash: &near_indexer::near_primitives::hash::CryptoHash,
     chunk_hash: &near_indexer::near_primitives::hash::CryptoHash,
     block_timestamp: u64,
     strict_mode: bool,
@@ -111,7 +110,7 @@ async fn find_tx_hashes_for_receipts(
     pool: &actix_diesel::Database<PgConnection>,
     mut receipts: Vec<near_indexer::near_primitives::views::ReceiptView>,
     strict_mode: bool,
-    block_hash: &str,
+    block_hash: &near_indexer::near_primitives::hash::CryptoHash,
     chunk_hash: &near_indexer::near_primitives::hash::CryptoHash,
 ) -> anyhow::Result<HashMap<String, String>> {
     let mut tx_hashes_for_receipts: HashMap<String, String> = HashMap::new();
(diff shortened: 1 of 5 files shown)


src/db_adapters/receipts.rs
```diff
@@ -13,16 +14,38 @@ use crate::schema;
...
+async fn store_chunk_receipts(
     pool: &actix_diesel::Database<PgConnection>,
     receipts: &[near_indexer::near_primitives::views::ReceiptView],
-    block_hash: &str,
+    block_hash: &near_indexer::near_primitives::hash::CryptoHash,
     chunk_hash: &near_indexer::near_primitives::hash::CryptoHash,
     block_timestamp: u64,
     strict_mode: bool,
 ) -> anyhow::Result<()> {
-    if receipts.is_empty() {
```
khorolets, line 23: Why did you remove this?

src/db_adapters/transactions.rs
```diff
@@ -1,22 +1,56 @@
...
+    if inserted_receipt_ids.len() == tried_to_insert_transactions_count {
+        return Ok(());
     }
 
-    let futures = shards
+    // https://github.com/near/near-indexer-for-explorer/issues/84
+    // TLDR: it's the hack to store transactions with collided hashes
+    // It should not happen, but unfortunately it did,
+    // we have ~10 such transactions in Mainnet for now
+    let transaction_hash_suffix = "_issue84_".to_owned() + &block_height.to_string();
```
khorolets, line 51: It looks more like a prefix

src/db_adapters/transactions.rs
```diff
@@ -25,33 +59,61 @@ pub(crate) async fn store_transactions(
                 chunk
                     .transactions
                     .iter()
-                    .collect::<Vec<&near_indexer::IndexerTransactionWithOutcome>>(),
+                    .enumerate()
```
khorolets, line 62: do we need this `enumerate` here? I don't see the use, could you point me where to look if I'm wrong?

khorolets (approved):

src/db_adapters/receipts.rs
telezhnaya: I moved this check into line 26 (`filter`)

src/db_adapters/transactions.rs
telezhnaya: I had to put it here because we need to pass indexes for collided transactions, they will not be sequential

src/main.rs
```diff
@@ -42,29 +42,30 @@ async fn handle_message(
...
+        streamer_message.block.header.timestamp,
+        strict_mode,
+    );
+
+    // We can process transactions and receipts in parallel
+    // because most of receipts depend on transactions from previous blocks,
+    // so we can save up some time here.
+    // In case of local receipts (they are stored in the same block with corresponding transaction),
+    // we hope retry logic will cover it fine
+    try_join!(transactions_future, receipts_future)?;
```
telezhnaya, line 68: @frol I think we want to put this change in a separate PR and test the performance separately

khorolets: I'd suggest including it here. We can deploy on separate testnet node to check how's the performance and revert it if necessary

src/db_adapters/transactions.rs
frol: We craft a prefix here to pass it into the store_chunk_transactions function below

frol (approved):

telezhnaya:
I want to save the reference here
https://github.com/near/near-indexer-for-explorer/blob/master/src/db_adapters/transactions.rs#L53-L57

```rust
let transaction_hash_suffix = "_issue84_".to_owned() + &block_height.to_string();
```

PR #213, opened 2021-12-14
Update ReadMe

DDeAlmeida:
Hi, after using ReadMe to manage the Guildnet explorer Indexer, i saw some modification required.
@frol if you agree with this modifications ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,44 +28,62 @@ Before you proceed, make sure you have the following software installed:
     On Debian/Ubuntu:
     
     ```bash
-    $ sudo apt install libpq-dev
+     sudo apt install libpq-dev
     ```
-
-
+ * Cargo dependency
+ * 
+   On Debian/Ubuntu:
+    
+    ```bash
+     sudo apt install cargo
+    ```
+    
 ### Prepare Database
 
-Setup PostgreSQL database, create a database with the regular tools, and note the connection string (database host, credentials, and the database name).
+Setup PostgreSQL database
+
+    ```bash
+     sudo apt-get install postgresql postgresql-contrib
+    ```
+    
+Create a database with the regular tools (Change user and db_name with yours)
+    ```bash
+     sudo su postgres
+     psql
+     create database db_name with owner=user;
+    ```
+Note the connection string (database host, credentials, and the database name).
 
 Clone this repository and open the project folder
 
 ```bash
-$ git clone git@github.com:near/near-indexer-for-explorer.git
-$ cd near-indexer-for-explorer
+ git clone https://github.com/near/near-indexer-for-explorer.git
+ cd near-indexer-for-explorer
 ```
 
 You need to provide database credentials in `.env` file like below (replace `user`, `password`, `host` and `db_name` with yours):
 
 ```bash
-$ echo "DATABASE_URL=postgres://user:password@host/db_name" > .env
+ echo "DATABASE_URL=postgres://user:password@host/db_name" > .env
 ```
 
 Then you need to apply migrations to create necessary database structure. For this you'll need `diesel-cli`, you can install it like so:
 
 
 ```bash
-$ cargo install diesel_cli --no-default-features --features "postgres"
+ cargo install diesel_cli --no-default-features --features "postgres"
 ```
 
 And apply migrations
 
 ```bash
-$ diesel migration run
+ diesel migration run
 ```
 
 ### Compile NEAR Indexer for Explorer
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,52 +28,70 @@ Before you proceed, make sure you have the following software installed:
...
-    $ sudo apt install libpq-dev
+     sudo apt install libpq-dev
     ```
+ * Cargo dependency
+ * 
+   On Debian/Ubuntu:
+    
+    ```bash
+     sudo apt install cargo
+    ```
```
frol, line 39: Installing Rust compiler is the first item on the list. Rustup is more flexible way to install and manage rustc/cargo versions

README.md
```diff
@@ -28,52 +28,70 @@ Before you proceed, make sure you have the following software installed:
     On Debian/Ubuntu:
     
     ```bash
-    $ sudo apt install libpq-dev
+     sudo apt install libpq-dev
```
frol, line 31: Well, it is a commit practice to use `$` to indicate that the command should be run on behalf of a regular user.

README.md
```diff
@@ -28,52 +28,70 @@ Before you proceed, make sure you have the following software installed:
...
+```   
 
-Setup PostgreSQL database, create a database with the regular tools, and note the connection string (database host, credentials, and the database name).
+Create a database with the regular tools (Change user and db_name with yours)
+    ```bash
+     sudo su postgres
+     psql
+     create database db_name with owner=user;
+    ```
+Note the connection string (database host, credentials, and the database name).
```
frol, line 55: Meh, setting up a database is a complicated topic, so I would avoid giving any instructions. Consider finding a good article out there and link it from this README

frol (changes requested):

README.md
DDeAlmeida: Yep but Cargo is mandatory for install diesel_cli
So he must be added to the list

README.md
DDeAlmeida: Oh ok, but this is a pain when copy/paste.
Sorry for this commit.

README.md
DDeAlmeida: Or we can just give it as a example. as you want

README.md
frol: cargo gets installed when you install Rustup. It is really hard to install Rust without cargo. Thus, there is no need for this extra step.

README.md
```diff
@@ -28,52 +28,70 @@ Before you proceed, make sure you have the following software installed:
...
 ```bash
-$ cargo install diesel_cli --no-default-features --features "postgres"
+ cargo install diesel_cli --no-default-features --features "postgres"
 ```
 
 And apply migrations
 
 ```bash
-$ diesel migration run
+ diesel migration run
```
OS-M, line 80: I think there should be "cargo diesel migration run". 
On my system (Ubuntu 21) "diesel migration run" ends up with "command not found" but my variant works fine

khorolets: Even after this command?

```
$ cargo install diesel_cli --no-default-features --features "postgres"
```

OS-M: Whoops my command doesnt work as well. 
`cargo install diesel_cli --no-default-features --features "postgres"` 
says 
`Ignored package `diesel_cli v1.4.1` is already installed, use --force to override`
but
`diesel migration run`
ends up with
`diesel: command not found`

OS-M: The problem was that i didnt have .cargo/bin in my PATH

OS-M:
I would also suggest removing 

> NB! According to changes in nearcore config generation we don't fill all the necessary fields in the config file. While this issue is open https://github.com/near/nearcore/issues/3156 you need to download config you want and replace the generated one manually.

Because mentioned issue is already closed

README.md
```diff
@@ -28,52 +28,70 @@ Before you proceed, make sure you have the following software installed:
...
+ cargo build --release
 ```
 
 ### Configure NEAR Indexer for Explorer
 
 To connect NEAR Indexer for Explorer to the specific chain you need to have necessary configs, you can generate it as follows:
 
 ```bash
-$ ./target/release/near-indexer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
+ ./target/release/near-indexer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
```
OS-M, line 94: As I see in current version the `./target/release/near-indexer` is renamed into `./target/release/indexer-explorer`

PR #224, opened 2022-01-13
feat(events): support FT events

telezhnaya:


src/db_adapters/assets/events.rs
```diff
@@ -0,0 +1,106 @@
...
+use diesel::PgConnection;
+use tracing::warn;
+
+use super::event_types;
+
+pub(crate) async fn store_events(
+    pool: &Database<PgConnection>,
+    streamer_message: &near_indexer::StreamerMessage,
+) -> anyhow::Result<()> {
+    for shard in &streamer_message.shards {
```
telezhnaya, line 12: futures

src/db_adapters/assets/events.rs
```diff
@@ -0,0 +1,106 @@
...
+                assets::event_types::NearEvent::Nep141(ft_event) => {
+                    assets::fungible_token_events::handle_ft_event(
+                        pool,
+                        shard,
+                        outcome,
+                        block_timestamp,
+                        &ft_event,
+                        &mut ft_index_in_shard,
+                    )
+                    .await?;
```
telezhnaya, line 62: future

telezhnaya: batch insert

telezhnaya: 500 at a time? 100? check implementation at transactions/receipts

telezhnaya (commented):
this is for me

src/db_adapters/assets/events.rs
frol: Just for the reference: https://github.com/near/near-indexer-for-explorer/pull/228

khorolets (approved):

telezhnaya:
Doc that we can share while talking about the feature
https://github.com/telezhnaya/near_contract_example/blob/main/events.md

telezhnaya:
@frol I fixed all the comments. If you want, you can have a look again

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -2,123 +2,72 @@ use actix_diesel::dsl::AsyncRunQueryDsl;
...
-                );
-                return None;
-            }
-        };
-
-        let nft_types::NearEvent::Nep171(nep171_event) = event;
-        Some(nep171_event)
-    });
-
+    let mut index_in_shard = 0;
```
frol, line 56: We can replace it with `events_with_outcomes.enumerate()` on the loop

src/db_adapters/assets/non_fungible_token_events.rs
```diff
@@ -2,123 +2,72 @@ use actix_diesel::dsl::AsyncRunQueryDsl;
...
     let mut nft_events = Vec::new();
```
frol, line 57:
```suggestion
    let mut nft_events = Vec::with_capacity(events_with_outcomes.len());
```

frol (approved):
Looks good to me, but see my comments and apply them across NFT/FT implementations

src/db_adapters/assets/non_fungible_token_events.rs
telezhnaya: No, the vector could be bigger and this will confuse the reader. The event can include more than one DB rows

src/db_adapters/assets/non_fungible_token_events.rs
telezhnaya: No, we can't, since we have more than one DB event inside each log

src/db_adapters/assets/non_fungible_token_events.rs
frol: Ouch, indeed, my bad.

src/db_adapters/assets/non_fungible_token_events.rs
frol: Oh, should we just use `ft_events.len()` instead of a separate variable `index_in_shard`?

telezhnaya: I'm not sure I got your idea.

We have 2 `for` cycles. We never calculate the sum of the sizes of the inner ones. We can add an additional cycle before this logic where we will calculate the capacity, but, IMO, that's overhead

PR #234, opened 2022-01-20
refactor: Add ReceiptsCache to decrease the number of unnecessary SELECT queries to the database

### LINKED ISSUE #229: Optimize parent receipts tracking
Until this point in time we maintained stateless indexing, where no information was passed from one block handler to another, but with the recent performance degradation I believe it is time to introduce some caching layer to potentially avoid SELECTing from the database in the most cases. Currently, the major contributor to the IO wait is the query that tries to find the parent receipt/transaction:

https://github.com/near/near-indexer-for-explorer/blob/1a0739806f6ef605dd6cf77744c2bdfbc07bd941/src/db_adapters/receipts.rs#L199-L224

If we maintain the cache of the parent receipt/transaction, and only fallback to SELECTing form the database when we don't have the information in our cache, we will gain quite some performance boost.

khorolets:
Resolves #229 

* I've added the `--debug` flag and the first `indexer_for_explorer` debug message to reason about how the cache is working, it should allow us to put some more logs if needed
* Add `ReceiptCache` which is a `HashMap` of accordance between Receipt ID and parent Transaction hash. 
  The working flow looks like this:
  * First we handle transactions and we store the pair `converted_into_receipt_id: transaction_hash` to the cache
  * When processing receipts we look into the cache to get a parent transaction hash for each of the receipts
    * [Fallback logic] If we don't observe parent transaction hash for specific receipts we are looking for them in the database
  * ExecutionOutcomes stage. We are interested in those outcomes that result in another receipt(s). We get all those receipts' ids and we take the parent transaction hash for the original receipt from the cache. Then we put `receipt_id: transaction_hash` for each of the new receipts.

### DIFF
--- a/src/db_adapters/execution_outcomes.rs
+++ b/src/db_adapters/execution_outcomes.rs
@@ -1,4 +1,5 @@
 use actix_diesel::dsl::AsyncRunQueryDsl;
+use cached::Cached;
 use diesel::PgConnection;
 use futures::future::try_join_all;
 
@@ -41,7 +42,7 @@ pub async fn store_execution_outcomes_for_chunk(
         // remove it from cache once found as it is not expected to observe the Receipt for
         // second time
         let parent_transaction_hash =
-            receipts_cache_lock.remove(outcome.execution_outcome.id.to_string().as_str());
+            receipts_cache_lock.cache_remove(&outcome.execution_outcome.id.to_string());
 
         let model = models::execution_outcomes::ExecutionOutcome::from_execution_outcome(
             &outcome.execution_outcome,
@@ -64,7 +65,7 @@ pub async fn store_execution_outcomes_for_chunk(
                     // could find their parents in cache
                     if let Some(transaction_hash) = &parent_transaction_hash {
                         receipts_cache_lock
-                            .insert(receipt_id.to_string(), transaction_hash.clone());
+                            .cache_set(receipt_id.to_string(), transaction_hash.clone());
                     }
                     models::execution_outcomes::ExecutionOutcomeReceipt {
                         executed_receipt_id: outcome.execution_outcome.id.to_string(),
--- a/src/db_adapters/execution_outcomes.rs
+++ b/src/db_adapters/execution_outcomes.rs
@@ -9,6 +9,7 @@ pub(crate) async fn store_execution_outcomes(
     pool: &actix_diesel::Database<PgConnection>,
     shards: &[near_indexer::IndexerShard],
     block_timestamp: u64,
+    receipts_cache: crate::ReceiptsCache,
 ) -> anyhow::Result<()> {
     let futures = shards.iter().map(|shard| {
         store_execution_outcomes_for_chunk(
@@ -16,6 +17,7 @@ pub(crate) async fn store_execution_outcomes(
             &shard.receipt_execution_outcomes,
             shard.shard_id,
             block_timestamp,
+            std::sync::Arc::clone(&receipts_cache),
         )
     });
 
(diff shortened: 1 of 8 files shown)


src/db_adapters/receipts.rs
```diff
@@ -112,8 +122,28 @@ async fn find_tx_hashes_for_receipts(
...
     let mut tx_hashes_for_receipts: HashMap<String, String> = HashMap::new();
+    //
+    tx_hashes_for_receipts.extend(
+        receipts_cache_lock
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone())),
+    );
+
+    // discard the Receipts already in cache from the attempts to search
+    receipts.retain(|r| !receipts_cache_lock.contains_key(r.receipt_id.to_string().as_str()));
```
frol, line 138: Instead of cloning the whole cache, move the retained elements (`retain` returns the removed elements and you don't need to clone them since it returns owned objects)

src/db_adapters/transactions.rs
```diff
@@ -102,14 +105,34 @@ async fn store_chunk_transactions(
     block_timestamp: u64,
     // hack for supporting duplicated transaction hashes. Empty for most of transactions
     transaction_hash_suffix: &str,
+    receipts_cache: crate::ReceiptsCache,
 ) -> anyhow::Result<()> {
+    let mut receipts_cache_lock = receipts_cache.lock().await;
```
frol, line 110: Same comment about releasing the lock ^^^

src/main.rs
```diff
@@ -28,11 +29,20 @@ const AGGREGATED: &str = "aggregated";
 const INTERVAL: std::time::Duration = std::time::Duration::from_millis(100);
 const MAX_DELAY_TIME: std::time::Duration = std::time::Duration::from_secs(120);
 
+// Introducing a simple cache for Receipts to find their parent Transactions without
+// touching the database
```
frol, line 33: Please, add details about the content since `HashMap<String, String>` is not descriptive enough.

khorolets: I was thinking to add type aliases
```
ReceiptId = String
TransactionHash = String
```
to be more explicit in the code. But I doubt it because we use Strings while real ids and hashes a `CryptoHash`. What do you think?

src/db_adapters/execution_outcomes.rs
```diff
@@ -28,11 +30,17 @@ pub async fn store_execution_outcomes_for_chunk(
...
+    receipts_cache: crate::ReceiptsCache,
 ) -> anyhow::Result<()> {
+    let mut receipts_cache_lock = receipts_cache.lock().await;
     let mut outcome_models: Vec<models::execution_outcomes::ExecutionOutcome> = vec![];
     let mut outcome_receipt_models: Vec<models::execution_outcomes::ExecutionOutcomeReceipt> =
         vec![];
     for (index_in_chunk, outcome) in execution_outcomes.iter().enumerate() {
+        // Trying to take the parent Transaction hash for the Receipt from ReceiptsCache
+        let parent_transaction_hash =
+            receipts_cache_lock.remove(outcome.execution_outcome.id.to_string().as_str());
```
telezhnaya, line 42: Are we sure whether the parent tx hash is no longer needed in the cache at this point?

src/main.rs
frol: Well, ReceiptId -> TransactionHash is an improvement, but it still does not communicate the fact that it is a relation for receipt to the transaction where that receipt is originated from.

src/db_adapters/execution_outcomes.rs
frol: @khorolets It is a fair question from @telezhnaya. While we know it is a correct invariant, we should communicate this explicitly.

khorolets: @telezhnaya I'm pretty sure. Are you thinking about something I could miss?

@frol do you want me to extend the comment about it or what?

telezhnaya: I had no idea of the bad scenario. That was just unclear for me from the brief reading of the code, why can we drop this cache entry without thinking that it could be useful again.

After the second (again brief) reading, I got the invariant this way: we will appear here once for each receipt. So we need each line in the cache only once; that's why we can drop it after the usage.

In this case, I see another potential problem: our cache can grow infinitely. It's a rare case and I'm sure we will not catch this point, but it's better not to rely on that.
Maybe we want to add removing the item from cache after 100 blocks passed and noone came for it.

src/db_adapters/receipts.rs
khorolets: Sorry, I didn't get what you propose

src/main.rs
khorolets: So don't you think such aliases might confuse more than help?

src/db_adapters/execution_outcomes.rs
khorolets: @telezhnaya I've replaced `HashMap` with `cached::SizedCache` to address the 
> our cache can grow infinitely

frol (approved):

telezhnaya (approved):

PR #243, opened 2022-01-26
Drop duplicated index, add index on action_receipt_actions

telezhnaya:
I also checked whether we have other duplicated indexes - no, it was the only one


migrations/2022-01-26-184200_index_action_receipt_actions/up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX CONCURRENTLY action_receipt_actions_receiver_and_timestamp_idx
```
frol, line 1: It seems we learned that diesel migration tool cannot really use `CONCURRENTLY` in the migrations since it does it wraps the migrations in transactions

frol (approved):
I am unblocking this PR, but address my comment before merging

telezhnaya:
@khorolets both of them are already on mainnet and testnet, so you don't need to apply them while deploying

PR #244, opened 2022-01-26
refactor: Improve ReceiptsCache to handle Data Receipts and some optimization

### LINKED ISSUE #242: "Looking for parent transaction hash in database for 1 receipts"
image

After a 10 minutes of indexer running I still see those warnings (we expect to only see them during a few first blocks, btw, we should make the message clear that it is fine to ignore those warnings during the indexer warm-up [or even better if we don't even print those warnings during the warm-up]).

Also, I see an increased number of

```
DatabaseError(ForeignKeyViolation, "insert or update on table \"receipts\" violates foreign key constraint \"tx_receipt_fk\"")
```

khorolets:
Resolves #242 
* Replace type alias `ReceiptIdString` with `ReceiptOrDataId` enum
* Cache parent transaction hash for Data Receipts (it is necessary to do in a different way)

### DIFF
--- a/src/db_adapters/receipts.rs
+++ b/src/db_adapters/receipts.rs
@@ -16,8 +16,6 @@ use tracing::{error, warn};
 use crate::models;
 use crate::schema;
 
-type ReceiptDataIdString = String;
-
 /// Saves receipts to database
 pub(crate) async fn store_receipts(
     pool: &actix_diesel::Database<PgConnection>,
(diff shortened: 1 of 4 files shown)


src/db_adapters/receipts.rs
```diff
@@ -69,31 +72,90 @@ async fn store_chunk_receipts(
...
-                skipping_receipt_ids.insert(r.receipt_id);
-                None
+            // We need to search for parent transaction hash in cache differently
+            // depending on the Receipt kind
+            // In case of Action Receipt we are looking for ReceiptId
+            // In case of Data Receipt we are looking for DataId
+            match r.receipt {
+                near_primitives::views::ReceiptEnumView::Action { .. } => {
+                    if let Some(transaction_hash) =
+                    tx_hashes_for_receipts.get(&crate::ReceiptOrDataId::ReceiptId(r.receipt_id))
```
frol, line 82: Instead of copy-pasting the whole body, you can match and craft the lookup-key:

```suggestion
            let receipt_or_data_id = match r.receipt {
                near_primitives::views::ReceiptEnumView::Action { .. } => {
                    crate::ReceiptOrDataId::ReceiptId(r.receipt_id)
                }
                near_primitives::views::ReceiptEnumView::Data { data_id, .. } => {
                    crate::ReceiptOrDataId::DataId(data_id)
                }
            };
            if let Some(transaction_hash) = tx_hashes_for_receipts.get(&receipt_or_data_id) {
```

src/db_adapters/receipts.rs
```diff
@@ -240,56 +344,68 @@ async fn find_tx_hashes_for_receipts(
...
-                ))
-                .load_async::<(crate::ReceiptIdString, crate::ParentTransactionHashString)>(pool),
-            10,
-            "Parent Transaction for Receipts were fetched".to_string(),
-            &receipts
-        )
-        .unwrap_or_default();
+                                })
+                                .map(|r| r.receipt_id.to_string())
+                                .collect::<Vec<String>>()
```
frol, line 372: Given that we introduced `ReceiptDataIdString` in this file, should we also re-introduce `ReceiptIdString` next to it or drop `ReceiptDataIdString`? Though, I see `ReceiptIdString` is also used in the `transactions.rs`, so we might as well just keep `ReceiptIdString` in main.rs

frol (approved):

src/db_adapters/receipts.rs
khorolets: No, I've dropped `ReceiptIdString` from `transactions.rs` and from the main. I'd vote for removing `ReceiptDataIdString` as well

frol (approved):

PR #251, opened 2022-02-07
fix(migrations): Fix migrations have the same timestamp

teke97:
Fixes near/near-indexer-for-explorer/#250


khorolets (approved):
Thank you!

PR #258, opened 2022-02-22
Mieubrisse/more dockerfile fixes

### LINKED ISSUE #180: Refactor Dockerfile to use entrypoint approach instead of cmd
In #177 a working Dockerfile was introduced. Thanks, @mieubrisse!

However, we'd like to consider refactoring the approach to use `ENTRYPOINT` instead of `CMD`

To allow users to use the published Docker image like:

```bash
$ docker run -it near/indexer-for-explorer --home-dir <path> run sync-from-latest
```

or to pass any other parameters and use it as a binary.

mieubrisse:
Fixes #180 

This will:
1) allow restarting a stopped indexer container by skipping the `diesel migration`/`near init` steps that we need to run before the indexer is usable (but shouldn't run if they're already done)
2) pull all the nasty logic that was in the Dockerfile into a clean script
3) use ENTRYPOINT so that any CMD args that get passed to the indexer will go directly to the indexer binary process


Dockerfile
```diff
@@ -63,15 +65,14 @@ RUN apt-get update -qq && apt-get install -y \
...
-# If the --store-genesis flag isn't set, the accounts in genesis won't get created in the DB which will lead to foreign key constraint violations
-# See https://github.com/near/near-indexer-for-explorer/issues/167
-CMD ./diesel migration run && \
-    ./indexer-explorer --home-dir /root/.near/localnet init ${BOOT_NODES:+--boot-nodes=${BOOT_NODES}} --chain-id localnet && \
-    sed -i 's/"tracked_shards": \[\]/"tracked_shards": \[0\]/' /root/.near/localnet/config.json && \
-    sed -i 's/"archive": false/"archive": true/' /root/.near/localnet/config.json && \
-    ./indexer-explorer --home-dir /root/.near/localnet run --store-genesis sync-from-latest
+
+COPY --from=build /usr/local/cargo/bin/diesel .
+# NOTE: We'd expect to need the diesel.toml here, but for some reason we don't
```
khorolets, line 71: `diesel.toml` contains settings that are used during the generation of `schema.rs` it's not involved in the migration process, your expectation here is incorrect :)

khorolets (approved):
LGTM

Dockerfile
mieubrisse: Removed!

PR #259, opened 2022-02-23
feat: update near version to 1.25.0-rc.2 and use tag not rev

Akagi201:


khorolets:
Using `rev` gives us more flexibility, sometimes we have updated on NEAR Indexer Framework in nearcore, which is not included in any tag yet. We're not going to migrate to `tag` sorry.

PR #266, opened 2022-04-05
Update README.md

OS-M:
December PR with editing README seems some kind of abandoned so i decided to make my own. 
I suggest replacing the `./target/release/near-indexer` with `./target/release/indexer-explorer` because is seems renamed in current version
And also as the https://github.com/nearprotocol/nearcore/issues/3156 is closed we dont need the NB! section anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ NEAR runs the indexer and maintains it for [NEAR Explorer](https://github.com/ne
 
 WARNING: We may evolve the data schemas, so make sure you follow the release notes of this repository.
 
-NOTE: Please, keep in mind that the access to the database is shared across everyone in the world, so it is better to make sure you limit the amount of queris and individual queries are efficient.
+NOTE: Please, keep in mind that the access to the database is shared across everyone in the world, so it is better to make sure you limit the amount of queries and individual queries are efficient.
 
 ## Self-hosting
 
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ $ cargo build --release
 To connect NEAR Indexer for Explorer to the specific chain you need to have necessary configs, you can generate it as follows:
 
 ```bash
-$ ./target/release/indexer-explorer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
+$ ./target/release/near-indexer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,17 +73,11 @@ $ cargo build --release
 To connect NEAR Indexer for Explorer to the specific chain you need to have necessary configs, you can generate it as follows:
 
 ```bash
-$ ./target/release/near-indexer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
+$ ./target/release/indexer-explorer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
 ```
 
 The above code will download the official genesis config and generate necessary configs. You can replace `testnet` in the command above to different network ID (`betanet`, `mainnet`).
 
-**NB!** According to changes in `nearcore` config generation we don't fill all the necessary fields in the config file.
```
khorolets, line 81: While I agree with removing this paragraph, I would like to keep the links to configs as they might be useful for future reference.

khorolets (changes requested):
Thank you for the PR! Small request from me

README.md
```diff
@@ -73,17 +73,11 @@ $ cargo build --release
 To connect NEAR Indexer for Explorer to the specific chain you need to have necessary configs, you can generate it as follows:
 
 ```bash
-$ ./target/release/near-indexer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
+$ ./target/release/indexer-explorer --home-dir ~/.near/testnet init --chain-id testnet --download-config --download-genesis
```
khorolets, line 76: Nice catch, but I'm not sure. @frol what do you think maybe it's better to change the name in `Cargo.toml` instead?

README.md
OS-M: Ok, but what is the way to keep them? Just keep the links or like "Here are the links to initial configs"?

README.md
frol: @khorolets I would rename it in Cargo.toml to `near-indexer-for-explorer` to match the name of the repo and avoid confusion

OS-M: Ok, i discarder there changes. Should I do one more PR with changing `Cargo.toml`?

frol: @OS-M Yes, please, create a separate PR with the change in Cargo.toml and README

README.md
frol: Yes, please, phrase it as these are the default config files that one could use just for the reference.

frol (approved):

frol:
@OS-M Thanks! 🚀

PR #271, opened 2022-04-20
docs: Fix broken links to backups

lebedev:
@khorolets I used the link containing all 4 options (test/mainnet, rpc/archival), which is more likely to be up-to-date compared to other links like from bootcamp articles.


khorolets (approved):
Great! Thanks!

PR #272, opened 2022-04-21
refacor: Upgrade nearcore to 1.26.0-rc.1

khorolets:
* Upgrade `nearcore` to 1.26.0-rc.1
* Bump version to 0.10.15
* Refactor `store_account_changes` according to breaking change in `near-indexer` 1.26.0


src/db_adapters/account_changes.rs
```diff
@@ -1,11 +1,25 @@
...
+) -> anyhow::Result<()> {
+    let futures = shards
+        .iter()
+        .map(|shard| store_state_changes(pool, &shard.state_changes, block_hash, block_timestamp));
+
+    try_join_all(futures).await.map(|_| ())
+}
+
+/// Saves state change related to account to database
+async fn store_state_changes(
```
telezhnaya, line 22: nit: in other files it's called `..._for_chunk`, let's call it `store_account_changes_for_chunk`?

khorolets: But it is for shard this time, `state_changes` are not in `chunk`, but I got you

Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
 name = "indexer-explorer"
-version = "0.10.14"
+version = "0.10.15"
 authors = ["Near Inc <hello@nearprotocol.com>"]
 edition = "2021"
 
 [dependencies]
-actix = "=0.11.0-beta.2"
+actix = "=0.13.0"
 actix-rt = "=2.2.0"  # remove it once actix is upgraded to 0.11+
```
frol, line 9: According to the comment, we should be able to remove this line (otherwise, update the comment)

Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
-actix-web = "=4.0.0-beta.6"
-actix-http = "=3.0.0-beta.6"
-actix-tls = "=3.0.0-beta.5"
-actix_derive = "=0.6.0-beta.1"
+# actix-web = "=4.0.0-beta.6"
+# actix-http = "=3.0.0-beta.6"
+# actix-tls = "=3.0.0-beta.5"
+# actix_derive = "=0.6.0-beta.1"
```
frol, line 13: If we don't need them, just remove

Cargo.lock
```diff
@@ -66,7 +90,7 @@ name = "actix-diesel"
 version = "0.3.1-alpha.0"
 source = "git+https://github.com/frol/actix-diesel?branch=actix-0.11-beta.2#7066a04ccce15b20940d926758837c99954765cc"
 dependencies = [
- "actix",
+ "actix 0.11.0-beta.2",
```
frol, line 93: Update actix-diesel commit hash to `3a001986c89dfabfc3c448d8bae28525101b4992` (I have just created that branch) to drop actix 0.11 from the dependencies

Cargo.lock
```diff
@@ -5421,6 +5557,49 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 
+[[package]]
+name = "windows-sys"
```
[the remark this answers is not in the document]
khorolets: I believe this is coming from `nearcore` 🤷‍♂️

Cargo.toml
```diff
@@ -39,6 +35,6 @@ tracing-subscriber = "0.2.4"
 uint = { version = "0.8.3", default-features = false }
 
 actix-diesel = { git = "https://github.com/frol/actix-diesel", branch = "actix-0.11-beta.2" }
```
frol, line 37:
```suggestion
actix-diesel = { git = "https://github.com/frol/actix-diesel", branch = "actix-0.13" }
```

Cargo.toml
```diff
@@ -1,23 +1,19 @@
...
 
 [dependencies]
-actix = "=0.11.0-beta.2"
-actix-rt = "=2.2.0"  # remove it once actix is upgraded to 0.11+
-actix-web = "=4.0.0-beta.6"
-actix-http = "=3.0.0-beta.6"
-actix-tls = "=3.0.0-beta.5"
-actix_derive = "=0.6.0-beta.1"
+actix = "=0.13.0"
+actix-rt = "=2.2.0"
```
frol, line 9: Well, you removed the comment, but do we actually still need to pin actix-rt version? I would imagine we can drop this line completely

src/db_adapters/account_changes.rs
frol: @khorolets you renamed not the function @telezhnaya was talking about. The top-level function should be called `store_account_changes` while the helper function that stores the account changes for a single shard: `store_account_changes_for_shard`

Cargo.lock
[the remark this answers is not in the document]
frol: It is coming from the upgraded version of one of the dependencies of actix (parking_lot), which migrated from winapi to windows-sys. Well, we should not worry about it.

telezhnaya (approved):

frol (approved):
See my last comment, but overall it is good to go

Cargo.toml
khorolets: I wanted to remove it, but we're using it

https://github.com/near/near-indexer-for-explorer/blob/3f1584b5a74236f89d89989f85db11c7d691a119/src/db_adapters/genesis.rs#L22-L25

I have unpinned the version

PR #293, opened 2022-08-05
Add login option for creating user

telezhnaya:
Without this option, users can't log in into db


khorolets (approved):
Don't forget to apply this please :)

telezhnaya:
There's no need to apply this, this is just a cheatsheet for creating new users :)

PR #297, opened 2022-08-16
fix: sha256 from DeployContract action is no base64 encoded anymore

khorolets:


frol (approved):

telezhnaya (commented):
If someone will run the indexer from genesis, it won't work

khorolets:
@telezhnaya yeah, thank you for the catch! I'll update the PR

Issue #304, opened 2022-09-10
Access Key matching the implicit accounts gets updated on each transfer


achernik:
Access Key matching the implicit accounts gets updated on each transfer, setting `last_update_block_height` to the block height of the transfer execution and `deleted_by_receipt_id` to null. 

Steps to reproduce: 
 1) Create an Implicit Account with a transfer
 2) Add another Full access key to that account
 3) Delete initial access key (the one matching the implicit account name)
 4) Make another transfer to the account

Result: 
 The initial access key in the DB looks like it hasn't been deleted, even though it has been.

frol:
Yeah, we worked under a bit different assumptions back in the days. We should not act like blockchain re-creates the access key with every transfer, yet we need to ensure that we create it with the first transfer (including the cases when the account is re-created after being deleted)

@achernik Thanks for raising it 👍 

@pkudinov I believe this is a good onboarding issue for Monday

frol:
Here is an account on testnet that exhibits this behaviour: https://explorer.testnet.near.org/accounts/cfb46d6579ca320a3dd78c7a3133b4a20bb6724302c413e792db7130bd740205

frol:
While the fix is implemented in #317, we will need to re-index the whole history of testnet and mainnet. To do so, I suggested disabling all the other tables, creating a new temporary table `access_keys_reindexed` alongside the existing tables, and re-indexing the whole history there, once all the data is there we can swap the tables and drop the old one.

cc @morgsmccauley @khorolets

PR #317, opened 2022-10-12
fix: Avoid recreating access key on transfer to implicit account

morgsmccauley:
This PR refactors the `access_keys` logic to be driven from [`StateChanges`](https://near-indexers.io/docs/data-flow-and-structures/structures/state_change#statechangewithcauseview) rather than [`ExecutionOutcomes`](https://near-indexers.io/docs/data-flow-and-structures/structures/execution_outcome), allowing us to remove the logic which indirectly creates an access key on every transfer to an implicit account. This means access keys will only be written when the implicit account is first created, and ignore subsequent transfers. Resolves https://github.com/near/near-indexer-for-explorer/issues/304.

Test plan:
- [x] verified access key updates still function as expected by running the `StateChange`/`ExecutionOutcome` logic simultaneously and comparing the proposed DB updates.
- [x] verified that access keys for implicit accounts are not overwritten on subsequent transfers (specifically `deleted_by_receipt_id` and `created_by_receipt_id`)

### DIFF
--- a/src/db_adapters/access_keys.rs
+++ b/src/db_adapters/access_keys.rs
@@ -2,7 +2,6 @@ use std::collections::HashMap;
 
 use actix_diesel::dsl::AsyncRunQueryDsl;
 use actix_diesel::Database;
-use bigdecimal::BigDecimal;
 use diesel::{ExpressionMethods, PgConnection, QueryDsl};
 use futures::try_join;
 use tracing::info;
@@ -25,22 +24,12 @@ pub(crate) async fn handle_access_keys(
     }
 
     let mut access_keys = HashMap::<(String, String), models::access_keys::AccessKey>::new();
-    let mut deleted_accounts = HashMap::<String, String>::new();
 
     for state_change in state_changes {
         if let near_primitives::views::StateChangeCauseView::ReceiptProcessing { receipt_hash } =
             state_change.cause
         {
             match &state_change.value {
-                near_primitives::views::StateChangeValueView::AccountDeletion { account_id } => {
-                    deleted_accounts.insert(account_id.to_string(), receipt_hash.to_string());
-                    access_keys
-                        .iter_mut()
-                        .filter(|((_, receiver_id), _)| receiver_id == &account_id.to_string())
-                        .for_each(|(_, access_key)| {
-                            access_key.deleted_by_receipt_id = Some(receipt_hash.to_string());
-                        });
-                }
                 near_primitives::views::StateChangeValueView::AccessKeyUpdate {
                     account_id,
                     public_key,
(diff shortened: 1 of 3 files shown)


src/db_adapters/access_keys.rs
```diff
@@ -15,107 +14,71 @@ use crate::{metrics, models};
...
-                                        &receipt.receipt_id,
-                                        block_height,
-                                    ),
-                                );
-                            }
-                        }
-                    }
-                    _ => continue,
+    for state_change in state_changes {
+        if let near_primitives::views::StateChangeCauseView::ReceiptProcessing { receipt_hash } =
```
morgsmccauley, line 31: Not sure if we should also be watching for any of the other enums which also contain `receipt_hash`?  (`ActionReceiptProcessingStarted`, `ActionReceiptGasReward`, or `PostponedReceipt`)

src/db_adapters/access_keys.rs
```diff
@@ -15,107 +14,71 @@ use crate::{metrics, models};
...
+            state_change.cause
+        {
+            match &state_change.value {
+                near_primitives::views::StateChangeValueView::AccountDeletion { account_id } => {
```
frol, line 35: I believe we don’t need a special case for account deletion as blockchain will generate state changes for removing all the account access keys individually, and thus you will catch them in the arm below. Also, this seems to mean that we can drop all the logic related to `deleted_accounts`

frol (commented):
Great start!

src/db_adapters/access_keys.rs
```diff
@@ -15,107 +14,71 @@ use crate::{metrics, models};
...
+                    deleted_accounts.insert(account_id.to_string(), receipt_hash.to_string());
+                    access_keys
+                        .iter_mut()
+                        .filter(|((_, receiver_id), _)| receiver_id == &account_id.to_string())
+                        .for_each(|(_, access_key)| {
+                            access_key.deleted_by_receipt_id = Some(receipt_hash.to_string());
+                        });
                 }
+                near_primitives::views::StateChangeValueView::AccessKeyUpdate {
```
morgsmccauley, line 44: This is triggered when an access key is created but the `Update` makes me wonder what else triggers it, is `insert`ing here going to be safe?

frol: Oh, good catch! The keys are updated with every transaction they are used to sign transactions, since nonce value gets updated. This means that we need to properly distinguish creation from these updates. Oops, we need to take some time to think through it. It seems there will be a need to make a hybrid approach of the previous one and the new one to catch access keys creation properly

frol: The solution is that once we observe the key update state change, check the receipt actions for the matching AddKey or Transfer (in case if the receiver account id is implicit, and if the updated key matches the implicit access key, `decode_hex(receipt.receiver_account_id) == updated_access_key.public_key.as_bytes()`).

frol: Disregard my previous comments. It will be TransactionProcessing event that updated the nonce, so we are good here with only handling ReceiptProcessing state changes. No changes needs to be made here

morgsmccauley: Awesome 👍🏽

frol (approved):

PR #321, opened 2022-11-22
chore: Reindex `access_keys`

morgsmccauley:
Creating a temporary branch which modifies indexer so that it only writes to `access_keys_reindexed`. This is so we can reindex the entire `access_keys` table so that the data is not affected by the bug described in #304.

As I am not that familiar with the codebase, I wanted to create this PR to get a second pair of eyes and ensure I am on the right track. I'll pull down this branch on the testnet/mainnet machines to build the binary, rather than patching it on the machine directly.

### DIFF
--- a/src/db_adapters/access_keys.rs
+++ b/src/db_adapters/access_keys.rs
@@ -82,20 +82,20 @@ pub(crate) async fn handle_access_keys(
 
     let update_access_keys_future = async {
         for value in access_keys_to_update {
-            let target = schema::access_keys::table
-                .filter(schema::access_keys::dsl::public_key.eq(value.public_key.clone()))
+            let target = schema::access_keys_reindexed::table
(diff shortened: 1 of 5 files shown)


frol:
@morgsmccauley LGTM. You should change the `--height 0` to match the block height of the genesis block (either check the `genesis.json` file or see the genesis height on Explorer for the respective network: https://explorer.near.org/stats)

morgsmccauley:
Thanks @frol, updated `--height` 👍🏽

morgsmccauley:
As we are in the process of migrating to the lake-based indexer, I'm going to wait until that is complete before reindexing this table. The bug fix for #304 still hasn't been deployed to production and doing so would be wasted work as they'll soon be shut down.

I'll raise another PR in the lake fork when ready to reindexing. I'll also lump this work in with "storing genesis" as that's also required for the reindex.

PR #322, opened 2022-11-30
doc: Cargo.toml fomatting guide

roshaans:
Cargo.toml formatting guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,10 +118,9 @@ The rationale behind this is that there are plenty of different `Account` types 
 Use specific versioning when specifying dependencies. It helps to mitigate potential security vulnerabilities and reason through code better.
 
 #### Choosing Secure Crates
-1. Use crates from reputable sources (crates.io)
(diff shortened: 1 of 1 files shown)


telezhnaya (changes requested):
Initially, I collected the info in a raw format. Could you please write it down with attention to the rules of English language and the common sense? Follow the way how the rest of the file is written.

Keep in mind that the rules are only about the dependencies, it's worth to mention that.

Also, `Cargo.toml` in this project does not follow these rules, you need to re-check and fix it.

Also, (it's not required but anyway) it would be nice if you read something about how `Cargo.toml` is organised in Rust and maybe add the rules or even edit the proposed ones.

telezhnaya (approved):
Thank you!

> Would you say it is considered best practice to pin down a specific version rather than using ^, ~, and * when specifying dependencies?

I haven't done a huge research here, but, intuitively, I don't think we need to insist on setting up the specific version. I even think it may be anti-pattern because with the specific version we will miss the non-breaking library updates that we may use for free.

Issue #256, opened 2022-02-18
Suggestion to make transactions and receipts adapters to sequential way


khorolets:
Recent testnet load with 200+ transactions and 300+ receipts were lagging. 

I really think this code causes the retry with a too heavy payload and we should try to avoid it

https://github.com/near/near-indexer-for-explorer/blob/08dfbe7d8f3a3e49ee42dd0720b444e8520cc147/src/main.rs#L86-L91

khorolets:
@telezhnaya what do you think?

/cc @frol

frol:
@khorolets Let's measure it on the specific block with and without concurrency and then make the final decision.

frol:
Just for the future investigation, the recent blocks that caused slowdowns on mainnet are around [#61617473](https://explorer.near.org/blocks/Snn61AxZbci6bkRUizmwcRCCw6v8tpnJgFhDydLiEJx)

PR #326, opened 2022-12-09
feat: Migrate to NEAR Lake

morgsmccauley:
This PR migrates the core indexing functionality to `near-lake-framework` by merging in the  [khorolets/near-indexer-for-explorer](https://github.com/khorolets/near-indexer-for-explorer) fork. Because the two branches have diverged, leading to many conflicts, it was easier to merge the fork in to a local branch, rather than creating a pull request from the fork directly.

We are in the process of phasing out the live nearcore based `indexer-for-explorer`s, replacing them with this lake fork, which makes this a good time to merge this back in. I've pushed the current `master` to [master-nearcore](https://github.com/near/near-indexer-for-explorer/tree/master-nearcore) so that the previous state is still available to build in case we need it, **can someone with admin access please make this branch protected**?

The code is mostly identical to [khorolets/near-indexer-for-explorer:master-lake](https://github.com/khorolets/near-indexer-for-explorer/tree/master-lake), I've commented inline where things differ.


.github/workflows/release.yml
```diff
@@ -0,0 +1,117 @@
+name: release
```
morgsmccauley, line 1: Binaries will now be built in Google Cloud Build, so maybe we can delete this now?

.github/workflows/test.yml
```diff
@@ -0,0 +1,85 @@
+name: ci
+on:
+  pull_request:
+    branches: [master]
+  push:
+    branches: [master]
```
morgsmccauley, line 6: Changed from `master-lake`

cross/x86_64-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:18.04
```
morgsmccauley, line 1: Not exactly sure what this and `Cross.toml` are. Seems `nearcore` related and they were also removed from this repo, maybe we can delete them now?

README.md
```diff
@@ -1,6 +1,6 @@
 # NEAR Indexer for Explorer
```
morgsmccauley, line 1: Removed all (NEAR Indexer for Explorer) "Lake" suffixes as it no longer needs to be differentiated from the lake fork, it _is_ the lake fork 😄

database/migrations/2022-06-07-000000_drop_action_receipt_actions_args_amount_index/down.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX action_receipt_actions_args_amount_idx ON action_receipt_actions((args->'args_json'->>'amount'))
```
morgsmccauley, line 1: Not exactly sure why these migrations existed in the lake fork and not here, should these still exist?

indexer/CHANGELOG.md
```diff
@@ -0,0 +1,272 @@
+# Changelog
+
+## 0.11.0
```
morgsmccauley, line 3: I've opted to keep the `near/near-indexer-for-explorer` `CHANGELOG.md` and add the lake migration as an entry on top. The migration seems more like a progression from the current indexer so this made more sense to me.

indexer/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "indexer-explorer-lake"
```
morgsmccauley, line 2: Haven't dropped `-lake` here, probably makes sense to keep it to easily differentiate?

cross/x86_64-unknown-linux-gnu/Dockerfile
khorolets: it's not `nearcore` related. This is used to build the binary. Previously adopted from https://github.com/near/near-lake-indexer (which is adopted from `near-cli-rs`, it's complicated :) )

database/migrations/2022-06-07-000000_drop_action_receipt_actions_args_amount_index/down.sql
khorolets: It might have been deleted in the original Indexer for Explorer while Lake fork existed. @telezhnaya do you know something about it?

indexer/Cargo.toml
khorolets: I guess we can drop the `-lake`. The changelog should give us a clear understanding since the `0.11` version Indexer for Explorer is based on Lake Framework.

README.md
```diff
@@ -15,11 +15,9 @@ NOTE: Please, keep in mind that the access to the database is shared across ever
...
-
 The final setup consists of the following components:
-* PostgreSQL database (you can run it locally or in the cloud), which can hold the whole history of the blockchain (as of January 2022, mainnet takes 1.1TB of data in PostgreSQL storage, and testnet takes 420GB)
-* NEAR Indexer for Explorer binary that operates as a regular NEAR Protocol peer-to-peer node, so you will operate it as any other [Archival Node in NEAR](https://docs.near.org/docs/develop/node/archival/hardware-archival)
+* PostgreSQL database (you can run it locally or in the cloud), which can hold the whole history of the blockchain (as of August 2022, mainnet takes 3TB of data in PostgreSQL storage, and testnet takes 1TB)
+* NEAR Indexer for Explorer binary that operates as a NEAR Lake Framework based indexer, it requires [AWS S3 credentials](https://near-indexers.io/tutorials/lake/credentials)
```
khorolets, line 20: There is an ongoing process of merging the docs from near-indexers.io to docs.near.org I believe it's important for us to try to give direct links to docs.near.org for those docs already merged.

```suggestion
* NEAR Indexer for Explorer binary that operates as a NEAR Lake Framework based indexer, it requires [AWS S3 credentials](https://docs.near.org/tutorials/indexer/credentials)
```

README.md
```diff
@@ -70,62 +75,39 @@ $ diesel migration run
...
-  - [mainnet config.json](https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/mainnet/config.json) 
+You can choose NEAR Indexer for Explorer start options:
+ - `from-latest` - start indexing blocks from the latest finalized block
+ - `from-interruption` - start indexing blocks from the block NEAR Indexer was interrupted last time but earlier for `<number_of_blocks>` if provided
+ - `from-block --height <block_height>` - start indexing blocks from the specific block height
 
-Configs for the specified network are in the `--home-dir` provided folder. We need to ensure that NEAR Indexer for Explorer follows
-all the necessary shards, so `"tracked_shards"` parameters in `~/.near/testnet/config.json` needs to be configured properly.
-For example, with a single shared network, you just add the shard #0 to the list:
+Unlike the original [NEAR Indexer for Explorer](https://github.com/near/near-indexer-for-explorer) you **can't** tell Indexer to store data from genesis (Accounts and Access Keys) by adding key `--store-genesis` to the `run` command. So please, ensure you took care about the genesis data in your database in order this indexer to work properly.
```
khorolets, line 87: Could you please create a tracking issue about the implementation of the store genesis feature and reflect the information in this paragraph instead?

README.md
```diff
@@ -70,62 +75,39 @@ $ diesel migration run
...
-trying to be stored. So Indexer keeps retrying to store the data until success. However if you're running Indexer not from the genesis it is possible that you
-really miss some of parent data and it'll be impossible to store child one, so you can disable strict mode for 1000 blocks to ensure you've passed the strong
-relation data area and you're running Indexer where it is impossible to loose any piece of data.
+NEAR Indexer for Explorer works in strict mode by default, but you can disable it. The strict mode means that every piece of data
+will be retried to store to database in case of error. Errors may occur when the parent piece of data is still processed but the child piece is already trying to be stored. So Indexer keeps retrying to store the data until success. However, if you're running Indexer not from the genesis it is possible that you really miss some of parent data and it'll be impossible to store a child one, so you can disable strict mode to ensure you've passed the strong relation data area and you're running Indexer where it is impossible to loose any piece of data.
```
khorolets, line 94: I wish some native speakers rephrase this paragraph to be understandable. I guess I've written it and I don't get a thing of it :(

README.md
```diff
@@ -70,62 +75,39 @@ $ diesel migration run
...
 By default NEAR Indexer for Explorer processes only a single block at a time. You can adjust this with the `--concurrency` argument (when the blocks are mostly empty, it is fine to go with as many as 100 blocks of concurrency).
 
 So final command to run NEAR Indexer for Explorer can look like:
 
 ```bash
-$ cargo run --release -- --home-dir ~/.near/testnet run --store-genesis --stream-while-syncing --non-strict-mode --concurrency 1 sync-from-latest
+$ ./target/release/indexer-explorer-lake \
+  --non-strict-mode \
+  --concurrency 1 \
+  from-latest
```
khorolets, line 110: This command misses the chain id. Could you add it in the proper place, please?

cross/x86_64-unknown-linux-gnu/Dockerfile
khorolets: Yeah, since we're going to build the binaries in GCB we should drop this file

indexer/CHANGELOG.md
khorolets: BTW, please note that we used to write the section `Breaking changes` describing what exactly is breaking. Could you try to write one here as well, please?

indexer/src/main.rs
```diff
@@ -0,0 +1,186 @@
...
+        streamer_message.block.header.timestamp,
+        strict_mode,
+        std::sync::Arc::clone(&receipts_cache),
+    );
+
+    // We can process transactions and receipts in parallel
+    // because most of receipts depend on transactions from previous blocks,
+    // so we can save up some time here.
+    // In case of local receipts (they are stored in the same block with corresponding transaction),
+    // we hope retry logic will cover it fine
```
khorolets, line 66: The approach described in this comment works. Though it constantly adds 1 retry and warning in the logs thus making logs unnecessarily noisy. I would vote to separate these processes as I don't think we are winning here much time. @telezhnaya what do you think?

indexer/src/main.rs
```diff
@@ -0,0 +1,186 @@
...
+    // Later we need to find the Receipt which is a parent to underlying Receipts.
+    // Receipt ID will of the child will be stored as key and parent Transaction hash/Receipt ID
+    // will be stored as a value
+    let receipts_cache: receipts_cache::ReceiptsCache =
+        std::sync::Arc::new(Mutex::new(SizedCache::with_size(100_000)));
+
+    let config: near_lake_framework::LakeConfig = opts.to_lake_config().await;
+    let (_, stream) = near_lake_framework::streamer(config);
+
+    tokio::spawn(async move {
```
khorolets, line 162: Could somebody remind me how it comes we use `[actix::main]` but `tokio::spawn`?

cc @frol @telezhnaya

khorolets (changes requested):
@morgsmccauley This is an awesome piece of work you've done! Thank you indeed!

FYI I've protected branches by `master-*` pattern (at least I hope so)

I've left a few comments here and there asking questions and requesting some changes. Overall is good, though I am not approving it yet, just because I want us to merge this PR with all the caution we can give.

@frol @telezhnaya I've cc'ed you on some questions and would be grateful if you review this PR too.

@morgsmccauley feel free to ping me when you believe this PR should be re-reviewed.

cross/x86_64-unknown-linux-gnu/Dockerfile
frol: Cross can be removed for near-lake-indexer as we basically support only x86_64-linux target (make sure CI building the binary release also does not use it). Cross is used for cross-target compilation of near-cli-rs binaries as we support ARM/x86/x86_64 Linux/macOS/Win there.

indexer/src/main.rs
frol: @khorolets It was recently introduced when metrics were added as either streaming or metrics need to be spawned (alternatively, they can be `select!`ed, but it does not make a lot of sense)

indexer/Cargo.toml
morgsmccauley: Good point, I'll remove it

README.md
morgsmccauley: Removed:
* `Dockerfile`
* `.dockerignore`
* `CHANGELOG.md` (in root dir)

database/migrations/2022-06-07-000000_drop_action_receipt_actions_args_amount_index/down.sql
morgsmccauley: Oh I know now - I somehow missed that the `migrations/` folder still existed in the root repository with these 3 migrations. That's why this was seen as an `git add` rather than a move. I've deleted the root `migrations/` and now it sees that they've been moved.

cross/x86_64-unknown-linux-gnu/Dockerfile
morgsmccauley: Thanks for the context 👍🏽 I've removed all `Cross` references

.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
...
+  # schedule:
+  # - cron: '00 01 * * *'
+jobs:
```
morgsmccauley, line 9: @khorolets I've removed `matrix` as it doesn't seem necessary now that we don't have `Cross`, let me know if this was incorrect and I'll add back in.

README.md
morgsmccauley: Done

README.md
```diff
@@ -70,62 +75,39 @@ $ diesel migration run
...
-Configs for the specified network are in the `--home-dir` provided folder. We need to ensure that NEAR Indexer for Explorer follows
-all the necessary shards, so `"tracked_shards"` parameters in `~/.near/testnet/config.json` needs to be configured properly.
-For example, with a single shared network, you just add the shard #0 to the list:
+Unlike the original [NEAR Indexer for Explorer](https://github.com/near/near-indexer-for-explorer) you **can't** tell Indexer to store data from genesis (Accounts and Access Keys) by adding key `--store-genesis` to the `run` command. So please, ensure you took care about the genesis data in your database in order this indexer to work properly.
 
 ```
-...
-"tracked_shards": [0],
-...
+TODO: Implement a tool to fill the database with the genesis data
```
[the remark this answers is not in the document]
morgsmccauley: Removed this, created #327 for tracking the addition of this feature, and updated to Readme referencing this issue.

indexer/CHANGELOG.md
morgsmccauley: Updated to the following:
```md
## 0.11.0

* Migrate to NEAR Lake Framework, `indexer-explorer` no longer runs a NEAR node and instead consumes block from [near-lake-indexer](https://github.com/near/near-lake-indexer).

### Breaking Changes

* Storing genesis file is no longer possible
* Command line arguments for starting the indexer have changed slightly, see [Readme](../README.md) for updated reference
* AWS credentials are now required for reading data from S3
```

Let me know if I've missed any breaking changes

README.md
morgsmccauley: I tried my best, how's this?

---

NEAR Indexer for Explorer works in strict mode by default. In strict mode, the Indexer will ensure parent data exists before storing children, infinitely retrying until this condition is met. This is necessary as a parent (i.e. `block`) may still be processing while a child (i.e. `receipt`) is ready to be stored. This scenario will likely occur if you have not stored the genesis file or do not have all data prior to the block you start indexing from. In this case, you can disable strict mode to store data prior to the block you are concerned about, and then re-enable it once you have passed this block.

To disable strict mode provide the following command arugment:

```
--non-strict-mode
```

morgsmccauley: I've already updated the `README.md` to include this, let me know if you're not happy with it.

Also including subheadings in this section to make things clearer.

.buildkite/pipeline.yml
```diff
@@ -1,13 +0,0 @@
-steps:
```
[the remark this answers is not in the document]
khorolets: Buildkite used to be faster than GitHub actions. I guess we can trade CI checks speed since we're going to build the binaries in GCB. @ecp88 what do you think?

.github/workflows/release.yml
khorolets: This is fine, though I don't think we want it to be built on each push on `master` it's better to build it on `tag` created.

indexer/CHANGELOG.md
khorolets: Please refer the issue to the point:

> Storing genesis file is no longer possible [issue_link]

just for the sake

README.md
khorolets: Yes! Thank you! Much better that I've written for sure :)

indexer/src/main.rs
khorolets: Maybe it is worth switching to `[tokio::main]`?

.github/workflows/test.yml
khorolets: Fully agree

khorolets (approved):
I am approving this PR to avoid blocking it, though I'd wait for @telezhnaya inputs before merging.

Also, I want to draw everybody's attention to questions:

1. Should we drop the Buildkite jobs since we can run checks via GitHub Actions and we're not going to build binaries with Buildkite anymore? (@ecp88 , @frol )
2. Is FOSSA still a thing and we should add its check to GH Actions? 
( @ecp88, @pkudinov ) 
3. Is it worth switching to `[tokio::main]` instead of `[actix::main]`? (@telezhnaya , @frol )

frol:
> Should we drop the Buildkite jobs since we can run checks via GitHub Actions and we're not going to build binaries with Buildkite anymore?

I don't know about the plans regarding Buildkite, but I would personally prefer Github Actions. Make your own judgement here.

> Is it worth switching to [tokio::main] instead of [actix::main]?

I would say it is a low-priority, and should only be done after the upgrade to near-lake-framework 0.6+ (after nearcore dependencies are upgraded to 0.15+ versions, which dropped actix as their leaf dependency), also, you will need to replace `actix-diesel` with `tokio-diesel`. Yet, after all of that the only observable benefits I can see are: (1) faster cold compilation, so expect some speed up on CI; (2) lower maintenance burden as actix used to cause some complications and we have to use a forked version fo actix-diesel now.

.github/workflows/release.yml
morgsmccauley: I'll remove the note about "every push..." and we can discuss the best approach for this when I create my deployment doc :)

indexer/CHANGELOG.md
morgsmccauley: Good call 👍🏽

morgsmccauley: Good call 👍🏽

indexer/src/main.rs
telezhnaya: Can we benchmark these 2 options and compare the results?
It could be done in the separate PR for sure

database/src/adapters/aggregated/circulating_supply.rs
```diff
@@ -23,7 +23,7 @@ pub(crate) async fn add_circulating_supply(
             }
             Err(async_error) => {
                 error!(
-                    target: crate::AGGREGATED,
+                    target: crate::EXPLORER_DATABASE,
```
telezhnaya, line 26: I suggest getting rid of using `explorer` word in indexer projects.
Historically, Explorer was the main and the only user for the DB, now it's not true.
What do you think about just `INDEXER_DATABASE`?

database/src/models/serializers.rs
```diff
@@ -82,17 +82,20 @@ pub(crate) fn extract_action_type_and_value_from_action_view(
...
+            // args_base64
             // See for reference https://github.com/near/near-indexer-for-explorer/issues/87
-            if let Ok(mut args_json) = serde_json::from_slice(&args) {
-                escape_json(&mut args_json);
-                arguments["args_json"] = args_json;
+            if let Ok(decoded_args) = base64::decode(args) {
+                if let Ok(mut args_json) = serde_json::from_slice(&decoded_args) {
+                    escape_json(&mut args_json);
+                    arguments["args_json"] = args_json;
+                }
```
telezhnaya, line 98: This change looks dangerous in this PR. I mean, all the other changes are all about changing the structure, library, or data source. This one changes the logic. 
If we have a bug in this, it would be hell to extract the bug from this huge PR.
Let's move this change to a separate PR?

database/src/retriable.rs
```diff
@@ -23,7 +23,7 @@ macro_rules! await_retry_or_panic {
                         })?
 
                         tracing::error!(
-                             target: crate::INDEXER_FOR_EXPLORER,
+                             target: crate::EXPLORER_DATABASE,
```
telezhnaya, line 26: nit:
All spell checkers always insist on `retryable`
Do we want to change the file name and the function name by any chance? :)

indexer/src/configs.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub async fn to_lake_config(&self) -> near_lake_framework::LakeConfig {
+        let config_builder = near_lake_framework::LakeConfigBuilder::default();
+
+        match &self.chain_id {
+            ChainId::Mainnet(_) => config_builder
+                .mainnet()
+                .start_block_height(get_start_block_height(self).await),
+            ChainId::Testnet(_) => config_builder
+                .testnet()
+                .start_block_height(get_start_block_height(self).await),
```
telezhnaya, line 87: this line could be lifted out of match clause

indexer/src/configs.rs
```diff
@@ -0,0 +1,168 @@
...
+        .build()
+        .expect("Failed to build LakeConfig")
+    }
+}
+
+async fn get_start_block_height(opts: &Opts) -> u64 {
+    match opts.start_options() {
+        StartOptions::FromBlock { height } => *height,
+        StartOptions::FromInterruption => {
+            let pool = models::establish_connection(&opts.database_url);
```
telezhnaya, line 98: We create the pool here and drop it at the end of the function.
Shouldn't we create it outside, in main?
I think it's better to pass to this function only `pool` and `start_options`

indexer/src/configs.rs
```diff
@@ -0,0 +1,168 @@
...
+    if std::env::var("ENABLE_JSON_LOGS").is_ok() {
+        subscriber.json().init();
+    } else {
+        subscriber.compact().init();
+    }
+
+    Ok(())
+}
+
+async fn final_block_height(opts: &Opts) -> u64 {
```
telezhnaya, line 159: It would be nice if the parameters help the reader to understand the logic without reading it.
Let's create rpc client only at main and pass it everywhere where it's needed? We almost there, it's the only exception

indexer/src/main.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use explorer_database::{adapters, models, receipts_cache};
+
+use crate::configs::Opts;
+
+mod configs;
+mod metrics;
+
+// Categories for logging
+const INDEXER_FOR_EXPLORER: &str = "indexer_for_explorer";
```
telezhnaya, line 17: Same, let's get rid of `explorer`?

indexer/src/main.rs
telezhnaya: Anyway, I suggest moving the logic as close to the master state as possible and work on any other ideas separately

indexer/src/main.rs
telezhnaya: Am I right that we use actix only because of diesel?
Anyway, I'm with @khorolets , let's use `[tokio::main]`

indexer/src/main.rs
```diff
@@ -0,0 +1,186 @@
...
+                handle_message(
+                    &pool,
+                    streamer_message,
+                    strict_mode,
+                    std::sync::Arc::clone(&receipts_cache),
+                )
+            })
+            .buffer_unordered(1usize);
+
+        while let Some(_handle_message) = handlers.next().await {}
```
telezhnaya, line 182: Major: we will silently die without even printing the error to logs

telezhnaya (changes requested):
Amazing work, thank you @morgsmccauley !

By the way, you've already migrated all the FT/NFT logic, as I see. We discussed that we need to keep the old indexer (node based) to collect FT/NFT stuff, but you've done everything we need, so we can kill the old instances, no blockers here from my side.
cc @khorolets

database/src/adapters/aggregated/circulating_supply.rs
khorolets: @telezhnaya but it is Explorer Database and we call it so. Let's keep it, at least as a tradition for this project. :)

database/src/models/serializers.rs
khorolets: The thing is, this project is not going to work without this change, because atm it uses an older version of `near-lake-framework` and that one returns this data in base64-encoded String, while newer version (nearcore as well returns raw `Vec<u8>`. So it should be here, and then it should be changed within the upgrade to `0.6.0` lake-framework

database/src/retriable.rs
khorolets: shame on me, we definitely should rename it if it is incorrect. @morgsmccauley could you please?

indexer/src/configs.rs
khorolets: This function is independent of the `main`. I did this intentionally, but now I see we can improve it, though it's out of the scope of this PR.

indexer/src/configs.rs
khorolets: I would suggest extracting this proposition to a separate issue for improvement. It is out of the scope of this PR and shouldn't block it

indexer/src/main.rs
khorolets: but it is the name of the project. Let's keep it as is, please :)

indexer/src/main.rs
khorolets: @telezhnaya I fully agree about separate PR, let's get back to it later. I believe I have created an issue about it a long time ago anyway

indexer/src/main.rs
khorolets: @telezhnaya yeah, I recall tokio-diesel didn't work (or stopped working), we can try to change it later, separately.

database/src/retriable.rs
morgsmccauley: Done

indexer/src/configs.rs
morgsmccauley: Good call, done.

database/src/adapters/aggregated/circulating_supply.rs
morgsmccauley: Agree with @khorolets here, we're writing to the `explorer` database so this naming makes sense?

indexer/src/main.rs
morgsmccauley: Added error handling 👍🏽. Decided to kill the entire process as that seems to be the consensus in https://github.com/near/near-indexer-events/pull/32#discussion_r1042698543

database/src/models/serializers.rs
morgsmccauley: Resolving as this change is required

morgsmccauley:
Thanks @telezhnaya @khorolets !!

I've addressed all comments which I could, these have been resolved.

To make things easier, I've summarised the unresolved discussions below. From the comments that @khorolets left I think the majority can be addressed in future PRs, but there are still two which require decisions:


|Discussion|Status|
|---|---|
|[Can we remove buildkite](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1045567018)|Needs decision|
|[Removing 'explorer' from names](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047107168)|Needs decision|
|[Can we separate storing transactions/receipts](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1044307250)|Separate PR|
|[Migrate from actix::main -> tokio::main and tokio-diesel](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1044308410)|Separate PR|
|[Creating a single PG pool](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047097965)|Separate PR|
|[Clearer parameters for final_block_height](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047105460)|Separate PR|

indexer/src/metrics.rs
```diff
@@ -0,0 +1,55 @@
...
+
+lazy_static! {
+    pub(crate) static ref LATEST_BLOCK_HEIGHT: IntGauge = try_create_int_gauge(
+        "indexer_explorer_latest_block_height",
+        "Height of last processed block"
+    )
+    .unwrap();
+    pub(crate) static ref BLOCK_COUNT: IntCounter =
+        try_create_int_counter("indexer_explorer_block_count", "Number of indexed blocks").unwrap();
+}
```
morgsmccauley, line 14: dropped the `-lake` prefix from the metrics since we've done the same everywhere else

telezhnaya (approved):

indexer/src/configs.rs
telezhnaya: Anyway, when we call it, we need to have `pool` ready

.buildkite/pipeline.yml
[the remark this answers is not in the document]
khorolets: Decision: we will remove buildkite once we completely shut down all the `nearcore`-based Indexer for Explorer nodes. Currently we need to update the Buildkite to work on `master-nearcore` branch to build binaries
cc @ecp88

database/src/adapters/aggregated/circulating_supply.rs
khorolets: Decision: we keep the Explorer mentions as this project and its database serves the needs of Explorer in the first place.

khorolets:
@morgsmccauley @telezhnaya @ecp88 

|Discussion|Outcome|
|---|---|
|[Can we remove buildkite](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1045567018)|Drop buildkite once we shut down nodes, [for now point it to `master-nearcore` branch](https://github.com/near/near-indexer-for-explorer/issues/332)|
|[Removing 'explorer' from names](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047077648)|We keep the Explorer mention in names|
|[Can we separate storing transactions/receipts](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1044307250)|[Separate issue](https://github.com/near/near-indexer-for-explorer/issues/256)|
|[Migrate from actix::main -> tokio::main and tokio-diesel](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1044308410)|[Separate issue](https://github.com/near/near-indexer-for-explorer/issues/330)|
|[Creating a single PG pool](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047097965)|[Separate issue](https://github.com/near/near-indexer-for-explorer/issues/331)|
|[Clearer parameters for final_block_height](https://github.com/near/near-indexer-for-explorer/pull/326#discussion_r1047105460)|[Separate issue](https://github.com/near/near-indexer-for-explorer/issues/331)|

morgsmccauley:
Thanks for creating all the relevant issues @khorolets 🙏🏽

PR #329, opened 2022-12-12
chore: Upgrade `nearcore` to `1.30.0`

morgsmccauley:
⚠️ This PR targets `master-nearcore` rather than `master` as we'll most likely be merging #326 soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.33
+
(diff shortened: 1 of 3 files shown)


khorolets (approved):
👍

khorolets:
@ecp88 I wonder whether Buildkite we used to have on `master` working for `master-nearcore`? Would you mind make it for, so we could build the binary the old way?

PR #335, opened 2023-01-08
fix: Add latest lockup contract hash to allow list

morgsmccauley:
The lockup contract was updated (https://github.com/near/core-contracts/pull/220) creating a new contract hash for lockup accounts. Currently we bail out of the circulating supply calculations when we encounter a hash we are not aware of. This PR adds the latest hash to the list of known contracts.


telezhnaya (approved):
Let's get rid of `bail` and add `log.info` for any new contract hash and date?

circulating-supply/src/lockup.rs
```diff
@@ -79,21 +80,17 @@ pub(super) async fn get_lockup_contract_state(
...
-        // The most fresh one
-        "4Pfw2RU6e35dUsHQQoFYfwX8KFFvSRNwMSNLXuSFHXrC" => Ok(false),
-        other => anyhow::bail!(
-            "Unable to recognise the version of contract {}, code hash {}",
-            account_id,
-            other
-        ),
+        "3kVY9qcVRoW3B5498SMX6R3rtSLiCdmBzKs7zcnzDJ7Q"
+        | "DiC9bKCqUHqoYqUXovAnqugiuntHWnM3cAc7KrgaHTu" => true,
+        _ => {
```
morgsmccauley, line 87: @telezhnaya - To avoid confusion, I've removed all contract hashes which return `false`. Now we just catch everything which isn't explicitly listed and assume it isn't buggy.

telezhnaya: If we have any issues with this part, we will definitely need the comments you dropped for debugging purposes. If it will be a task for the new engineer, they will suffer a lot.

I suggest leaving a comment with an explanation about all known lockup hashes (including the one you've met).

morgsmccauley: Good point 👍🏽 I've added the contract hashes and comments back in, including the new hash.

telezhnaya (approved):

PR #339, opened 2023-01-30
GCP storage

mhalambek:


indexer/src/configs.rs
```diff
@@ -81,6 +97,21 @@ impl Opts {
         match &self.chain_id {
             ChainId::Mainnet(_) => config_builder.mainnet(),
             ChainId::Testnet(_) => config_builder.testnet(),
+            ChainId::Custom(_) => {
+                println!("aws: region {}", self.region);
```
khorolets, line 101:
```suggestion
                tracing::debug!(
                    target: crate::INDEXER_FOR_EXPLORER,
                    "aws: region {}", 
                    self.region,
                );
```

indexer/Cargo.toml
```diff
@@ -35,3 +37,4 @@ near-jsonrpc-client = "0.4.0-beta.0"
 near-lake-framework = "0.5.2"
 
 explorer-database = { path = "../database" }
+http = "0.2.8"
```
khorolets, line 40: Would you mind moving it before the `itertools` declarations, please? This is the place for path-dependencies

indexer/src/configs.rs
```diff
@@ -34,6 +36,15 @@ pub(crate) struct Opts {
     /// Port to enable metrics/health service
     #[clap(long, short, env, default_value_t = 3030)]
     pub port: u16,
+    /// S3 bucket name
+    #[clap(long)]
+    pub bucket: String,
```
khorolets, line 41: This makes these parameters required which is not a desired behavior for us. It's actually required only when the `custom` subcommand is fired.

I see two possible options here:

1. Make these 'Option<String>` and add corresponding checks when the `Custom` is provided and print relevant errors about missing parameters.
2. Make a copy of the `StartOptions` and require these parameters there.

I guess the second variant is preferable.

khorolets (changes requested):
@mhalambek thank you for the PR. I see the point and fully support this feature (I was thinking about it for a long time). 

In order to merge this feature into the Indexer for Explorer I need you to change a few things here and there, I've left the comments. Feel free to ping me if you have questions.

khorolets:
@mhalambek hey there! FYI we have implemented the `store-genesis` feature (#340).

Do you have the capacity and desire to pick this PR again and finish it?

PR #340, opened 2023-01-31
Store genesis accounts

mhalambek:


khorolets:
Closing it since we've implemented it the way we planned in the #342

PR #342, opened 2023-02-01
feat: Add `from-genesis` subcommand

### LINKED ISSUE #327: Store genesis file with Lake based indexer
Soon, we will migrate to the Lake base indexer for explorer (#326). As Lake does not run a near node, we most likely do not have access to the genesis file. It is for this reason that this functionality was not implemented.

While we are not sure on the approach for storing the genesis file for a Lake based indexer, this feature should be implemented _some how_ .

morgsmccauley:
This PR adds the `from-genesis` subcommand which does the following:
1. Downloads the genesis file from S3 and stores the accounts/access keys defined in it
2. Starts indexing from the genesis block

This does not delete the genesis file after downloading it, this job will be left to the executor.

Majority of this code has been copied from the `master-nearcore` version of storing genesis records, the obvious difference being that we now need to download the genesis file. I've commented inline where other things differ. 

Closes #327


database/src/adapters/genesis.rs
```diff
@@ -0,0 +1,59 @@
+use actix_diesel::Database;
```
morgsmccauley, line 1: The [`master-nearcore` version](https://github.com/near/near-indexer-for-explorer/blob/master-nearcore/src/db_adapters/genesis.rs#L22) of this file included code to 'execute an async body on a specified actix runtime'. I'm not sure if this is still necessary as the excluded in still seemed to work.

indexer/src/configs.rs
```diff
@@ -113,6 +122,9 @@ async fn get_start_block_height(opts: &Opts) -> u64 {
             last_indexed_block
         }
         StartOptions::FromLatest => final_block_height(opts).await,
+        // Since NEAR Lake stores blocks in ascending order, using 0 here forces
+        // near-lake-framework to start from the first block, i.e. genesis
+        StartOptions::FromGenesis => 0,
```
morgsmccauley, line 127: I'm still not sure whether we should rely on this implicit behaviour, but this approach was much cleaner than getting the genesis block height here or overriding it in `main.rs`.

indexer/src/main.rs
```diff
@@ -156,16 +179,21 @@ async fn main() -> anyhow::Result<()> {
     let receipts_cache: receipts_cache::ReceiptsCache =
         std::sync::Arc::new(Mutex::new(SizedCache::with_size(100_000)));
 
-    let config: near_lake_framework::LakeConfig = opts.to_lake_config().await;
-    let (_, stream) = near_lake_framework::streamer(config);
```
morgsmccauley, line 160: Moving initialisation to after genesis processing

indexer/src/configs.rs
khorolets: IMO totally fine.

indexer/src/main.rs
```diff
@@ -156,16 +179,21 @@ async fn main() -> anyhow::Result<()> {
...
 
+    tokio::spawn(metrics::init_server(opts.port).expect("Failed to start metrics server"));
+
+    if opts.start_options() == &StartOptions::FromGenesis {
+        let genesis_file_path = download_genesis_file(&opts).await?;
+        adapters::genesis::store_genesis_records(&pool, genesis_file_path).await?;
+    }
+
+    let config: near_lake_framework::LakeConfig = opts.to_lake_config().await;
+    let (_, stream) = near_lake_framework::streamer(config);
```
khorolets, line 195: While you are here could you replace the `_` with the `sender` and handle the possible errors like it is suggested in the Lake Framework README, please?

ref: https://github.com/near/near-lake-framework-rs/blob/main/README.md?plain=1#L36-L53

indexer/src/configs.rs
```diff
@@ -72,6 +74,13 @@ impl Opts {
             ChainId::Testnet(_) => "https://rpc.testnet.near.org",
         }
     }
+
+    pub fn genesis_file_url(&self) -> &str {
+        match self.chain_id {
+            ChainId::Mainnet(_) => "https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/mainnet/genesis.json",
+            ChainId::Testnet(_) => "https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/testnet/genesis.json",
```
khorolets, line 81: I think we need somehow handle the case that @mhalambek is handling in #339 by providing the `Custom` chain id. But I am not sure how would be the best way, though.

database/src/adapters/genesis.rs
khorolets: While it still works doesn't mean "in still" works faster than async. Perhaps it is worth keeping.

@frol do you have any thoughts on that, please?

indexer/src/configs.rs
khorolets: Perhaps we might handle that separately after PR #339 is finished.

khorolets (approved):
LGTM overall! I've left a few comments. I don't want to be a blocker here, so I approve it while expecting you to solve my remarks. Thanks!

khorolets:
@ecp88 I remember we talked about disabling buildkite jobs for all the branches except `master-nearcore`. Do you have any updates on that?

database/src/adapters/genesis.rs
```diff
@@ -0,0 +1,59 @@
...
+) -> anyhow::Result<()> {
+    let mut accounts_to_store: Vec<crate::models::accounts::Account> = vec![];
+    let mut access_keys_to_store: Vec<crate::models::access_keys::AccessKey> = vec![];
+
+    let genesis = Genesis::from_file(genesis_file_path, GenesisValidationMode::Full);
+
+    let genesis_height = genesis.config.genesis_height;
+
+    for record in genesis.records.0 {
+        if accounts_to_store.len() == 5_000 {
```
telezhnaya, line 20: Why it should be exactly 5000?

database/src/adapters/genesis.rs
```diff
@@ -0,0 +1,59 @@
...
+                    &access_key,
+                    genesis_height,
+                ));
+            }
+            _ => {}
+        };
+    }
+
+    store_accounts_from_genesis(pool, accounts_to_store).await?;
+    store_access_keys_from_genesis(pool, access_keys_to_store).await?;
```
telezhnaya, line 56: Do we want to save the genesis block itself? For mainnet, it's 9820210, and we currently do not show it in Explorer

database/src/adapters/genesis.rs
khorolets: This figure was found experimentally for the bulk size

database/src/adapters/genesis.rs
khorolets: I remember doing something in this regard long time ago. The work has been done on the indexer framework side. 
Perhaps we had to index the genesis block and haven't done it yet. Don't remember it. But anyway this feature stores the data from genesis and has nothing to do with the indexing of very first block

database/src/adapters/genesis.rs
telezhnaya: Why do we ignore the data if the chunk is smaller? What if it's the last chunk?

database/src/adapters/genesis.rs
telezhnaya: As I understand this, 9820210 is the genesis block

database/src/adapters/genesis.rs
frol: @morgsmccauley It was essential back then because otherwise we halt all the async threads on the blocking read and blocking JSON parsing of genesis file (on testnet it is 5GB JSON file and takes 5-10 minutes to get through it). In the case of nearcore, we blocked p2p communication and other tasks, and then it did not feel healthy. Given that we don't do anything fancy in background, we can get away with blocking the whole async runtime here, but I would still consider extracting it into a separate thread just to avoid unnecessary surprises if `store_genesis_records` would be used under different circumstances assuming it would cooperate (won't block runtime).

Another problem your code did not account for is that loading 5GB JSON from disk and then parsing it requires quite a bit of RAM, which is why [`for_each_record` was used](https://github.com/near/near-indexer-for-explorer/blob/master-nearcore/src/db_adapters/genesis.rs#L63), that processes the file in a streaming manner instead of loading everything into RAM. That `for_each_record` expecting sync callback function was the main reason for `block_on` helper existence as we needed to run async code that pushes data into the database from this sync function that we wanted to run in a separate thread to avoid blocking the async runtime where `store_genesis_records` was executed (sync `main` -> async `main` -> async `store_genesis_records` -> sync `for_each_record` handler -> async `store_accounts_from_genesis`)

database/src/adapters/genesis.rs
morgsmccauley: We don't ignore it, accounts/access keys are also stored after looping through all genesis records (L55/56)

indexer/src/configs.rs
morgsmccauley: Let's wait til either of these PR's are merged, then we can address compatibility between the them in the one which isn't merged :)

database/src/adapters/genesis.rs
morgsmccauley: Thanks for the explanation @frol, I didn't even consider how this code would behave when dealing with a large genesis file.

> I would still consider extracting it into a separate thread just to avoid unnecessary surprises

Given how long a large genesis file will take to download, and then process, I wonder if blocking would be beneficial here. If we do this in parallel with streaming/storing block data, we'd likely encounter data which is dependant on genesis which would result in continuous retries until either the data is processed by the genesis thread or the number of retries is reached. I'm not sure how likely this is, but it seems pointless to start streaming blocks if we're going to run in to these blockers anyway. Was this a problem with the `nearcore` indexer?

Regardless I think there are two things I can do now to avoid the RAM issue:
1. Stream genesis file download to disk rather loading the entire response in memory and then writing to disk
2. Stream parsing of genesis file/records (after scanning `nearcore` code I'm not sure this is possible but will continue looking in to it)

frol: > I wonder if blocking would be beneficial here.

Well, if so, it is better to block explicitly with `.await` instead of implicitly relying on the fact that the whole async runtime thread will get blocked here. I think we should consider this currently blocking implementation as buggy since async functions must cooperate (do not block for too long).

> If we do this in parallel with streaming/storing block data

We should not do that in parallel with streaming the block data before we finish processing genesis exactly due to the named reasons.

> Was this a problem with the nearcore indexer?

Nope. The problem with the nearcore indexer was that we wanted nearcore to be able to do its work - connect to other nodes, sync and validate blocks, serve RPC, etc. If we block async runtime with a long-running blocking operation, none of the above can happen.

> Regardless I think there are two things I can do now to avoid the RAM issue:

An ideal implementation would use async read that would be directly receiving the data from the network or reading from disk (if users downloaded the genesis file manually), and parsing that on the fly, so we won't even need to store it on disk if we can read from the network. Yet, that might be too complicated and we might even fall back on asking users to manually download the genesis file (e.g. provide `curl` / `wget` command if there is no file on disk) as downloading big genesis files would ideally report some progress...

morgsmccauley: Had a chat with @frol offline, here's a summary:

**Problems:**
* This function is the combination of sync/async functionality as Frol mentions, specifically that `Genesis::from_file()` is sync while `store_genesis_records()` and the rest of it's implementation is async
* The genesis file is first loaded in to memory and then each record is iterated, this requires a lot of RAM with large genesis files, i.e. testnet.

**Next steps:**
1. Stream download of genesis file and show progress in logs
2. Make genesis file validation configurable via CLI flag to allow faster processing of genesis records for when we know the genesis file is valid
3. Avoid re-downloading genesis file if it already exists on disk. This also plays nicely in to the flag introduced in (2.)
4. Use `Genesis::new_with_path()` to avoid loading entire file in to memory. Use `for_each_record()` to stream genesis records. As this is synchronous, add the old blocking logic back in so that it works with the async diesel code

database/src/adapters/genesis.rs
telezhnaya: > the first block in mainnet S3 is 9820210

Great, I haven't checked that and I was thinking it could be related

indexer/src/main.rs
```diff
@@ -139,6 +141,27 @@ async fn handle_message(
...
+    };
+
+    info!(
+        target: INDEXER_FOR_EXPLORER,
+        "Downloading {} genesis file", chain
+    );
+
+    let resp = reqwest::get(opts.genesis_file_url()).await?;
+
+    let file_path = format!("{}-genesis.json", chain);
```
khorolets, line 157: I don't think it's necessary to have a prefix with a chain id for the file. Usually, a single instance of an indexer is used for one of the chains.
The only thing we should care about is ensuring the file is stored next to the indexer binary.

Another variant is to follow `nearcore` approach when the configs and files required by the instance is stored in under the `--home-dir`.

indexer/src/main.rs
```diff
@@ -165,16 +188,21 @@ async fn main() -> anyhow::Result<()> {
...
     tracing::info!(
         target: INDEXER_FOR_EXPLORER,
         "Starting Indexer for Explorer (lake)...",
     );
 
+    tokio::spawn(metrics::init_server(opts.port).expect("Failed to start metrics server"));
+
+    if opts.start_options() == &StartOptions::FromGenesis {
+        let genesis_file_path = download_genesis_file(&opts).await?;
+        adapters::genesis::store_genesis_records(&pool, genesis_file_path).await?;
```
khorolets, line 200: Would be lovely to drop a log message about the start of storing the genesis records, just for clarity.

khorolets (commented):
Let me know when you finished and I can start a review (disregard my current approval, please)

kampleforth:
when is this slated to merge? I'm having trouble with the `from-block` subcommand, it always fails with database query errors

morgsmccauley:
@kampleforth I am working on this now, you can expect it soon

database/src/adapters/genesis.rs
morgsmccauley: @frol @khorolets I have implemented most of the steps above, but currently, am still using `Genesis::from_file()`. I attempted to use `Genesis::new_with_path()`; however, it seems that it's still not possible to prevent loading the entire file into memory.

The reason for this is because `new_with_path<P: AsRef<Path>>(config: GenesisConfig, records_file: P)` requires `GenesisConfig` as an argument and constructing it involves [reading the entire file](https://github.com/near/nearcore/blob/0c673d4470307236aa8b3a17d51822c6361eb7a3/core/chain-configs/src/genesis_config.rs#L276), exactly what we are trying to avoid.

I tried passing in `GenesisConfig::default()`, as we may be able to work around not having `config.genesis_height`, but this doesn't work as `GenesisValidationMode` is not configurable and [defaults to](https://github.com/near/nearcore/blob/0c673d4470307236aa8b3a17d51822c6361eb7a3/core/chain-configs/src/genesis_config.rs#L465) `Full`. Consequently, validation fails as the defaults are insufficient. Additionally, all records are streamed/validated, but not loaded in to memory completely, potentially causing some unwanted delays.

I'm not sure what the best path forward is here. Maybe I have misunderstood some of this. Would appreciate some guidance here from you two! :)

indexer/src/main.rs
```diff
@@ -139,6 +141,40 @@ async fn handle_message(
...
+            info!("Using existing genesis file: {}", file_path);
+        }
+        Err(_) => {
+            let mut file = std::fs::File::create(&file_path)?;
+            let mut downloaded = 0;
+            let mut stream = res.bytes_stream();
+            while let Some(chunk) = stream.next().await {
+                let chunk = chunk?;
+                downloaded = std::cmp::min(downloaded + (chunk.len() as u64), total_size);
+                info!(
```
morgsmccauley, line 162: This was the simplest approach to showing download progress but it is very noisy. I considered using `indicatif::ProgressBar` but this doesn't integrate well with `tracing` so we wouldn't see progress in e.g. Cloud Run. I'll keep looking in to this, maybe we can log every ~10 chunks or so.

khorolets: First, we're trying to call tracing macros in the full name `tracing::info!` in the indexer code-base.

So I tried to run the code. Instead of any download progress, it just keeps the silence 

```
Mar 24 14:18:58.347  INFO indexer_for_explorer:Starting Indexer for Explorer (lake)...
Mar 24 14:18:58.348  INFO indexer_for_explorer:Starting metrics server on http://0.0.0.0:3030

```

Once I've added a proper `target: INDEXER_FOR_EXPLORER` it started to print something. So, please, ensure it works out of the box. Yeah, agree it is too noisy.

As for your concern about CloudRun, we usually don't run `from-genesis` there. 

I can come up with a few variants on how to handle it:

1. Use some fancy human-friendly progress bar there and just deal with the fact CloudRun won't like it (we can ignore it)
2. Depending on how exactly CloudRun won't like the fancy progress bar we can:
    - Add info-logs with dedicated target `info!(target: "genesis-download", ...)` and expect CloudRun to catch it since we run it using `RUST_LOG=debug`. Thus using both fancy progress bars and info-logs users are not subscribed to by default.
    - Or add an optional switch via CLI args between a human-readable progress bar and noisy naive info-logs

I don't have any strong preferences here. Though I believe everyone can survive even those noisy download logs.

indexer/src/main.rs
morgsmccauley: > I don't think it's necessary to have a prefix with a chain id for the file. Usually, a single instance of an indexer is used for one of the chains.

I've removed the chain ID prefix 👍🏽 

> The only thing we should care about is ensuring the file is stored next to the indexer binary

Updated so `genesis.json` is stored next to the executable. During local dev that means storing in`./target/{debug|release}`.

indexer/src/main.rs
morgsmccauley: Done

database/src/adapters/genesis.rs
morgsmccauley: Discussed with @khorolets offline. Unfortunately, it doesn't look like there is any way around the problems outlined above without some work on the `nearcore` side. We've therefore decided to:
1. Stick with loading the entire file in to memory
2. Add a note to `README.md` outlining the RAM concern

Note for (1.) I've decided to use `Genesis::new_with_path()` rather than `Genesis::from_file()`. While it does load the entire file for parsing `GenesisConfig`, it won't load all `GenesisRecord`s in to memory, which allows us to stream the records.

indexer/src/main.rs
morgsmccauley: > Once I've added a proper target: INDEXER_FOR_EXPLORER it started to print something.

Woops, I've added `target` 👍🏽 

> As for your concern about CloudRun, we usually don't run from-genesis there.

This might be the case now, but with the constant re-genesis on `betanet` I'd expect that we would run the `betanet` cloud run instance `from-genesis` permanently?

> Though I believe everyone can survive even those noisy download logs.

Yeah perhaps you're right, in a local scenario `from-genesis` won't be that frequent, and we reuse the existing genesis files so it will only download at least once. Regardless, I've updated the logic so that it logs every 100KB which has reduced the noise a bit.

Happy to keep iterating on this if you think the current solution isn't desirable :)

morgsmccauley:
@khorolets this is ready for re-review now :)

PR #343, opened 2023-02-02
fix: improve performance of assets__* by changing PKs

telezhnaya:
These changes have a major issue: when someone runs the project on the empty DB, the final state will not be accurate. I have no idea how to force users to read the migrations and run something manually.

I can uncomment everything and drop `CONCURRENTLY`: in this case, people with empty DB will be happy, while the others will have few hours blocking operation, which is unacceptable.

What do you guys think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ And apply migrations
 $ cd database && diesel migration run
 ```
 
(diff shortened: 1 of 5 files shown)


eduohe:
> These changes have a major issue: when someone runs the project on the empty DB, the final state will not be accurate. I have no idea how to force users to read the migrations and run something manually.
> 
> I can uncomment everything and drop `CONCURRENTLY`: in this case, people with empty DB will be happy, while the others will have few hours blocking operation, which is unacceptable.
> 
> What do you guys think?

We can't block users, this is a change to increase performance, maybe adding a recommendation in the docs and/or release notes can help.

eduohe (approved):

PR #344, opened 2023-02-03
fix: rewrite last assets migrations, update readme

telezhnaya:
I realised that we cheat on our users in the previous PR #343. 
It's been a long time when we add the migrations which is impossible to apply on the full DB: we just know about that and apply these changes manually with `CONCURRENTLY` option enabled.

I fixed the last 2 transactions so that it works well on the empty DB;
I also added the explanations to the README, adding the instruction how to apply the new migrations on the full DB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,9 @@ And apply migrations
 $ cd database && diesel migration run
 ```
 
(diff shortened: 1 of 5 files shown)


khorolets (approved):
LGTM

PR #347, opened 2023-02-07
fix: handle streamer result properly

telezhnaya:
copy-pasted from https://github.com/near/near-lake-framework-rs#example


khorolets (approved):
LGTM! Thanks for the PR!

PR #350, opened 2023-02-14
add the hack to enable the features

telezhnaya:
@eduohe I've tested `start-from-interruption` logic, it works fine, but I haven't checked the features part since it's not fully ready yet


khorolets (approved):
Nice

eduohe (approved):

PR #354, opened 2023-02-15
feat: rewrite start-from-interruption logic

telezhnaya:
With these changes, we:
- fail early if we have DB error
- start 500 blocks before the last block to be sure we haven't missed anything
- start from the beginning if the DB is empty

### DIFF
--- a/indexer/src/configs.rs
+++ b/indexer/src/configs.rs
@@ -101,8 +101,8 @@ async fn get_start_block_height(opts: &Opts) -> u64 {
                 last_indexed_block.saturating_sub(500)
             } else {
                 tracing::warn!(
-                    target: "indexer-for-explorer",
-                    "It seems the database is empty. Will start from the beginning",
+                    target: crate::INDEXER_FOR_EXPLORER,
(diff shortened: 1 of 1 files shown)


indexer/src/configs.rs
```diff
@@ -93,24 +93,19 @@ async fn get_start_block_height(opts: &Opts) -> u64 {
...
-            last_indexed_block
+            let last_indexed_block = adapters::blocks::latest_block_height(&pool)
+                .await
+                .expect("Failed to get last indexer block from Database");
+            if let Some(last_indexed_block) = last_indexed_block {
+                // -500 helps us to be sure we haven't missed anything
+                last_indexed_block.saturating_sub(500)
+            } else {
+                tracing::warn!(
+                    target: "indexer-for-explorer",
```
morgsmccauley, line 104:
```suggestion
                    target: crate::INDEXER_FOR_EXPLORER,
```

morgsmccauley (approved):
Looks good, just one tiny thing

PR #358, opened 2023-03-01
refactor: rearrange methods in tx_actions, receipt_actions

telezhnaya:
I've made some refactoring before I'll introduce meta tx support logic
1. get rid of `std::sync::Arc::clone` because it looks scary to me, we can just use `.clone()` instead, the same method is used
2. split the methods. I'll add more logic to some of them further
3. parallelize receipts tables inserts, it should work a little faster with this change
4. fix a small bug in `transaction_actions`: tx_hash for duplicated transaction will point out to its parent instead of the first original tx

### DIFF
--- a/database/src/adapters/execution_outcomes.rs
+++ b/database/src/adapters/execution_outcomes.rs
@@ -10,7 +10,7 @@ pub async fn store_execution_outcomes(
     pool: &actix_diesel::Database<PgConnection>,
     shards: &[near_indexer_primitives::IndexerShard],
     block_timestamp: u64,
-    receipts_cache: crate::receipts_cache::ReceiptsCache,
+    receipts_cache_arc: crate::receipts_cache::ReceiptsCacheArc,
 ) -> anyhow::Result<()> {
     let futures = shards.iter().map(|shard| {
         store_execution_outcomes_for_chunk(
@@ -18,7 +18,7 @@ pub async fn store_execution_outcomes(
             &shard.receipt_execution_outcomes,
             shard.shard_id,
             block_timestamp,
-            receipts_cache.clone(),
+            receipts_cache_arc.clone(),
         )
     });
 
@@ -31,12 +31,12 @@ pub async fn store_execution_outcomes_for_chunk(
     execution_outcomes: &[near_indexer_primitives::IndexerExecutionOutcomeWithReceipt],
     shard_id: near_indexer_primitives::types::ShardId,
     block_timestamp: u64,
-    receipts_cache: crate::receipts_cache::ReceiptsCache,
+    receipts_cache_arc: crate::receipts_cache::ReceiptsCacheArc,
 ) -> anyhow::Result<()> {
     let mut outcome_models: Vec<models::execution_outcomes::ExecutionOutcome> = vec![];
     let mut outcome_receipt_models: Vec<models::execution_outcomes::ExecutionOutcomeReceipt> =
         vec![];
-    let mut receipts_cache_lock = receipts_cache.lock().await;
+    let mut receipts_cache_lock = receipts_cache_arc.lock().await;
     for (index_in_chunk, outcome) in execution_outcomes.iter().enumerate() {
         // Trying to take the parent Transaction hash for the Receipt from ReceiptsCache
         // remove it from cache once found as it is not expected to observe the Receipt for
--- a/database/src/adapters/execution_outcomes.rs
+++ b/database/src/adapters/execution_outcomes.rs
@@ -18,7 +18,7 @@ pub async fn store_execution_outcomes(
             &shard.receipt_execution_outcomes,
             shard.shard_id,
             block_timestamp,
-            std::sync::Arc::clone(&receipts_cache),
+            receipts_cache.clone(),
         )
     });
 
--- a/database/src/adapters/receipts.rs
+++ b/database/src/adapters/receipts.rs
@@ -21,7 +21,7 @@ pub async fn store_receipts(
     block_hash: &near_indexer_primitives::CryptoHash,
     block_timestamp: u64,
     strict_mode: bool,
-    receipts_cache: crate::receipts_cache::ReceiptsCache,
+    receipts_cache_arc: crate::receipts_cache::ReceiptsCacheArc,
 ) -> anyhow::Result<()> {
     let futures = shards
         .iter()
@@ -35,7 +35,7 @@ pub async fn store_receipts(
                 &chunk.header.chunk_hash,
                 block_timestamp,
                 strict_mode,
-                receipts_cache.clone(),
+                receipts_cache_arc.clone(),
             )
         });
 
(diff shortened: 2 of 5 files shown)


database/src/adapters/execution_outcomes.rs
```diff
@@ -18,7 +18,7 @@ pub async fn store_execution_outcomes(
             &shard.receipt_execution_outcomes,
             shard.shard_id,
             block_timestamp,
-            std::sync::Arc::clone(&receipts_cache),
+            receipts_cache.clone(),
```
khorolets, line 21: The reason behind using "scary" `std::sync::Arc::clone()` was explicitness. And I really don't want to get rid of it.

With the "scary" call a reader understands that it's an `Arc` clone at a glance. With your change, a reader wouldn't be able to get it immediately without diving into the codebase.

Please, revert this change.

database/src/adapters/receipts.rs
```diff
@@ -146,9 +146,9 @@ async fn store_chunk_receipts(
         });
 
     let process_receipt_actions_future =
-        store_receipt_actions(pool, action_receipts, block_timestamp);
+        store_receipt_actions(pool, &action_receipts, block_timestamp);
 
-    let process_receipt_data_future = store_receipt_data(pool, data_receipts);
+    let process_receipt_data_future = store_data_receipts(pool, &data_receipts);
```
khorolets, line 151: This is really nice renaming!

khorolets (changes requested):
Amazing job! Great improvements! 

However, I am asking you to revert the changes made to the "scary" calls. After that we're good to merge :)

database/src/adapters/execution_outcomes.rs
telezhnaya: I disagree.
In master, we already have 3 places where we use `receipts_cache.clone()`: it's just impossible to remember that we have to use an explicit form

If you go to the definition of ReceiptsCache, you see it's Arc, it's not diving deep in my opinion.

And, btw, why it's important for the reader to know it's an Arc clone?

khorolets: > In master, we already have 3 places where we use receipts_cache.clone(): it's just impossible to remember that we have to use an explicit form

That highlights someone's mistakes (most likely mine). I would appreciate it if you update them to explicit calls as well :)

> If you go to the definition of ReceiptsCache, you see it's Arc, it's not diving deep in my opinion.

Can't argue with that it would be obvious if a reader had seen it.

> And, btw, why it's important for the reader to know it's an Arc clone?

I don't know, please accept it as my personal psychological issues. The project is somehow complicated, we are transferring the maintenance to another person and I would prefer we keep caring about the possible readers.

telezhnaya: As a reader and non-native Rust speaker, when I see `std::sync::Arc::clone`, I think "wow, I'd never thought I have to use so unintuitive form, I'll never write valid Rust code"
Moreover, while developing Indexer for the last 2 years, I've read the definition of `std::sync::Arc::clone` several times just because it disrupts me from reading, and I need to check why we have to use this. The truth is: we don't have to, this logic is going out of the box.

The only situation when you need to know that it's an Arc clone and the strong pointers will be created under the hood, is the moment when you've already dived super deep. In this case, the developer will not need this code, they are already much deeper than we are.

database/src/adapters/receipts.rs
```diff
@@ -486,14 +486,43 @@ async fn save_receipts(
...
         .iter()
         .filter_map(|receipt| models::ActionReceipt::try_from(*receipt).ok())
         .collect();
+    crate::await_retry_or_panic!(
+        diesel::insert_into(schema::action_receipts::table)
+            .values(receipt_actions.clone())
+            .on_conflict_do_nothing()
+            .execute_async(pool),
+        10,
+        "ReceiptActions were stored in database".to_string(),
```
morgsmccauley, line 515: https://github.com/near/near-indexer-for-explorer/blob/30f3d407c1ac8a76e3659466b974dc4060965521/database/src/retryable.rs#L2-L3

This is actually an error message, maybe this wording makes more sense? Same for all other cases below

```suggestion
        "Failed to store ReceiptActions in database".to_string(),
```

database/src/adapters/receipts.rs
morgsmccauley: agree!

database/src/adapters/receipts.rs
```diff
@@ -486,14 +486,43 @@ async fn save_receipts(
 
 async fn store_receipt_actions(
```
morgsmccauley, line 487:
```suggestion
async fn store_action_receipts(
```
I wonder if we can do a similar renaming here, like you have with data receipts. This conflicts with the new method you already created, so maybe we can just merge that logic in here?

Not sure which is better but I like the idea of avoiding use of `receipt_action` and `receipt_data` for naming action/data receipts. I easily confuse them with the data/actions within the receipts themselves 😅

database/src/adapters/execution_outcomes.rs
morgsmccauley: I think each of your points are valid:
1. Why be explicit when `Arc` provides the `.clone()` syntax/functionality out of the box
2. It's not immediately obvious that this is an `Arc`, within an IDE it's not too hard to figure out, but it's still mental overhead that is better avoided.

How about this for a middle ground solution:
* Rename `ReceiptsCache` type -> `ReceiptsCacheArc`
* Rename all `receipt_cache` variables -> `receipts_cache_arc`

This way the explicitness is tied to the variable/type itself and we don't need to remember to use `std::sync::Arc::clone` everywhere, instead we can just use `.clone()`

Thoughts? :)

morgsmccauley (commented):
Good job on this! Left a few comments :)

database/src/adapters/execution_outcomes.rs
telezhnaya: I like this!
@khorolets what do you think?

database/src/adapters/receipts.rs
telezhnaya: Agree, I saw many places where the naming/phrases could be improved, just decided not to dig too deep into this :)

database/src/adapters/execution_outcomes.rs
khorolets: Yeah I can live with that. Thanks @morgsmccauley

database/src/adapters/receipts.rs
khorolets: 😅 offtopic: I wish someone could explain to me why did I do such bad naming in the first place. There definitely was a reason I don't recall.

database/src/adapters/receipts.rs
khorolets: I believe this particular function literally stores the `actions` of the `receipt`, so this function name is actually reflecting what it does. And the parent of all that actions is an `ActionReceipt`. 

Thus I think your suggestion is not going to work.

khorolets (approved):

PR #361, opened 2023-03-02
Meta TX support

telezhnaya:


database/src/adapters/assets/fungible_token_events.rs
```diff
@@ -75,7 +75,7 @@ fn compose_ft_db_events(
                         event_memo: mint_event
                             .memo
                             .clone()
-                            .unwrap_or_else(|| "".to_string())
+                            .unwrap_or_default()
```
khorolets, line 78: I looked at the docs for [`unwrap_or_default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default) and now I wonder what is the default for the `memo`. I don't see a change to the structure behind `memo` with impl of a default. Did you do it earlier?

database/Cargo.toml
```diff
@@ -22,9 +22,9 @@ tracing = "0.1.13"
 tokio = { version = "1.1", features = ["sync", "time"] }
 
 actix-diesel = { git = "https://github.com/frol/actix-diesel", rev = "3a001986c89dfabfc3c448d8bae28525101b4992" }
-near-indexer-primitives = "0.14"
-near-primitives = "0.14"
-near-crypto = "0.14"
+near-indexer-primitives = "0.16.0"
```
khorolets, line 25: FYI we expect `0.16.1` to be released with re-exporting of the entire `near-primitives` to be done before `1.32.0` nearcore release.

database/src/adapters/assets/fungible_token_events.rs
telezhnaya: Discussed this on the call
`.unwrap_or_default()` is the idea from clippy, it takes default value for each type
Empty string in our case

indexer/src/configs.rs
```diff
@@ -81,7 +86,9 @@ impl Opts {
         match &self.chain_id {
             ChainId::Mainnet(_) => config_builder.mainnet(),
             ChainId::Testnet(_) => config_builder.testnet(),
+            ChainId::Betanet(_) => config_builder.betanet(),
         }
+        .s3_region_name("us-east-1") // hack because of the wrong region in lake framework
```
telezhnaya, line 91: We need to drop this before we merge it to master

database/Cargo.toml
khorolets: It seems they have forgotten about it, so we're flying without 0.16.1

khorolets (approved):
Great job @telezhnaya! Please drop that hack for the `betanet` region.

I believe we need to update the changelog before merging

indexer/Cargo.toml
```diff
@@ -31,8 +31,8 @@ tracing = "0.1.13"
 tracing-subscriber = "0.2.4"
 uint = { version = "0.8.3", default-features = false }
 
-near-jsonrpc-client = "0.4.0-beta.0"
-near-lake-framework = "0.5.2"
+near-jsonrpc-client = "0.5.0"
+near-lake-framework = "0.7.0"
```
khorolets, line 35: Don't forget to update the Lake Framework
```suggestion
near-lake-framework = "0.7.1"
```

morgsmccauley (approved):
Good job 👏🏽 LGTM!

PR #362, opened 2023-03-03
feat: Add the support of betanet (added with lake-framework-0.7.0)

khorolets:
This PR adds the support of `betanet` chain introduced in `near-lake-framework` 0.7.0

I believe most users won't need it, but it's a step in our work in creating an environment for international testing.

**Note**, this PR is created against `meta_tx` branch and wants to be merged into #361


telezhnaya (approved):
Thank you!

PR #365, opened 2023-03-08
fix: Action's args serialization to base64

khorolets:
Some time ago since `near-primitives` and `near-indexer-primitives` of version `0.15.0` arguments in some of the structures on the `nearcore` became pure `Vec<u8>` instead of base64-encoded `String`.

I remember a few PRs with the relevant changes on our side:
- #297 
- #306 

Which were overwritten back since we couldn't upgrade from `0.14.0` version of the primitives here #326 

Since the #361 we upgraded to `0.16.0` version of the primitives and now we have the same issue that we had with the attempt to migrate to `*-primitives-0.15.0`.

With this PR I am applying the changes necessary to handle the `Vec<u8>` data properly before storing it in the database.

We must release `0.12.1` in order for all of this to work before deploying and mark `0.12.0` as broken (consider even deleting the `0.12.0` one)

### DIFF
--- a/database/src/models/serializers.rs
+++ b/database/src/models/serializers.rs
@@ -68,11 +68,7 @@ pub(crate) fn extract_action_type_and_value_from_action_view(
         ActionView::CreateAccount => (ActionKind::CreateAccount, json!({})),
         ActionView::DeployContract { code } => (
             ActionKind::DeployContract,
-            json!({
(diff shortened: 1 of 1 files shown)


telezhnaya (approved):
Thank you for such a detailed explanation and fast fix!

PR #370, opened 2023-04-10
fix: Remove decoding of lockup contract as its now handled by nearcore

morgsmccauley:
Version `0.15` of `near-jsonrpc-primitives` returns values which are already `base64` decoded, making our manual decoding superfluous. When upgrading `0.16` we forgot to remove it, removing it now to fix the circulating supply computation.


eduohe (approved):
thanks for fixing this

PR #373, opened 2023-04-20
Hide tables to a separate schema

telezhnaya:
I tried to restrict selecting from 2 tables (see the full story [here](https://github.com/near/near-indexer-for-explorer/discussions/351)). I came to the conclusion that managed Postgres does not allow this.
That's why I created a `deprecated` schema, the users will not have access to it by default.

My plan is to move these tables to `deprecated` schema, hide them from all the users and see what happens.
We will still write the data to that tables.
If we break something important, we will grant access to the tables and consider how to fix it.
After a month of silence, we can drop them completely.

More tech details on what's going on here:
- Added migration
- Updated `schema.patch` so that it can generate the correct `schema.rs` file, it was a little broken
- Add the explanation of how diesel works with several schemas (TLDR: it does not work with several schemas, I hacked it)

I will also need to update the `features_enabled` branch. Waiting for your review first

### DIFF
--- a/database/diesel.toml
+++ b/database/diesel.toml
@@ -3,6 +3,8 @@
 
 [print_schema]
 file = "src/schema.rs"
+# Temporary add this if you need to update something in deprecated schema (and comment filter section)
+# schema = "deprecated"
 patch_file = "src/schema.patch"
 import_types = ["diesel::sql_types::*", "crate::models::enums::*"]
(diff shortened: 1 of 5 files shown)


morgsmccauley (approved):

khorolets (approved):
Excellent idea with a `deprecated` schema 👍

